PR #2, opened 2024-05-23
change dpop-based to dpop-inspired

bc-pi:
nitpicking on words already but "based" has a meaning to me that I don't think we want to convey here.


bc-pi:
Eventually I think/hope that "DPoP" won't even appear in this draft.

yaronf:
Yes, "inspired" is more correct.

PR #3, opened 2024-05-24
Intro and high level architecture

yaronf:


draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -52,34 +52,83 @@ TODO Abstract
...
+      v                            v
++------------+               +------------+
+|            |               |            |
+|  Identity  |               |    PDP     |
+|   Server   |               | (optional) |
+|            |               |            |
++------------+               +------------+
 ~~~
 
+The Identity Server provisions credentials to each of the workloads. At least Workload A (and possibly both) must be provisioned
```
arndt-s, line 103: Reading this makes me feel like the arrow in above picture from Workload B <-> Identity Server is missing. Feel free to resolve if it's left out on purpose as it's optional.

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -52,34 +52,83 @@ TODO Abstract
 
 # Introduction
 
-TODO Introduction
+This document defines authentication and authorization in the context of interaction between two workloads.
+This is the core component of the WIMSE architecture {{?I-D.ietf-wimse-arch}}.
+Assume that Service A needs to call Service B. For simplicity, this document focuses on REST services,
```
arndt-s, line 57: I think we should stick to one term - either `workload` or `service`. Using these two interchangeably may confuse readers.

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -52,34 +52,83 @@ TODO Abstract
...
+We propose to address them differently.
+
+* Many use cases have various middleboxes inserted between pairs of services, resulting in a transport layer
+that is not end-to-end encrypted. We propose to address these use cases by protecting the REST messages at the application
+level ({{app-level}}).
+
+* The other commonly deployed architecture has a mutual-TLS connection between each pair of services. This setup
+can be addressed by a simpler solution ({{mutual-tls}}).
+
+It is an explicit goal of this protocol that a service deployment, and in fact a single call chain,
```
arndt-s, line 71:
```suggestion
It is an explicit goal of this protocol that a service deployment can include both architectures across a multi chain call.
```
or do you indeed mean mTLS + TLS terminating middlebox? If that's the case I propose to rephrase the example to something like "TLS terminating middlebox + mTLS"

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -52,34 +52,83 @@ TODO Abstract
...
+|   Server   |               | (optional) |
+|            |               |            |
++------------+               +------------+
 ~~~
 
+The Identity Server provisions credentials to each of the workloads. At least Workload A (and possibly both) must be provisioned
+with a credential before the call can proceed. Details of communication with the Identity Server are out of scope
+of this document, however we do describe the credential received by the workload.
+
+PEP is a Policy Enforcement Point, the component that allows the call to go through or blocks it. PDP is an optional
```
arndt-s, line 107: This is very opinionated about authorization in my opinion. I propose to mention authz but not include it into the high level architecture. 
What do others think?

arndt-s (commented):
Please feel free to just close my comments if they are too nit picking at this stage.

draft-sheffer-wimse-s2s-protocol.md
yaronf: I omitted it for visual clarity more than anything else. I guess I should add it.

There's a related discussion that we haven't had yet on how much we want to protect the API response. If we decide to sign it then Service B absolutely needs an identity key.

draft-sheffer-wimse-s2s-protocol.md
yaronf: Agreed.

draft-sheffer-wimse-s2s-protocol.md
yaronf: I meant it exactly like you wrote it: a single multi chain call including both architectures. (Except that maybe the word "architecture" is a bit off here.)

draft-sheffer-wimse-s2s-protocol.md
yaronf: In what way is this opinionated in your opinion? :-) All I'm saying is that Service B is expected to make/enforce an authz decision. I'm just using common terminology from the policy world to do it.

draft-sheffer-wimse-s2s-protocol.md
yaronf: On second thoughts, even in the group charter we called it "service to service". I will let others chime in.

jsalowey: I'm in favor of switching to workload to workload vs service.  I feel that the workload term is more generic and its what we are using elsewhere.  I believe the way we are using the they are synonymous.

jsalowey (approved):
I'm in favor of merging this, I think we can work out the details later.

draft-sheffer-wimse-s2s-protocol.md
arndt-s: Let's discuss on Thursday. Resolving this for now to not block the PR.

arndt-s (approved):

PR #5, opened 2024-05-29
change REST to HTTP

bc-pi:
There's nothing necessarily RESTful about this nor should it be limited to REST. Let's just not use the term. 

https://htmx.org/essays/how-did-rest-come-to-mean-the-opposite-of-rest/ is a fun semi-related read, if you have or can find the time.

### DIFF
--- a/draft-sheffer-wimse-s2s-protocol.md
+++ b/draft-sheffer-wimse-s2s-protocol.md
@@ -56,7 +56,7 @@ TODO Abstract
 
 This document defines authentication and authorization in the context of interaction between two workloads.
 This is the core component of the WIMSE architecture {{?I-D.ietf-wimse-arch}}.
-Assume that Service A needs to call Service B. For simplicity, this document focuses on REST services,
+Assume that Service A needs to call Service B. For simplicity, this document focuses on HTTP services,
 and the service-to-service call consists of a single HTTP request and its response.
 We define the credentials that both services should possess and how they are used to protect the HTTP exchange.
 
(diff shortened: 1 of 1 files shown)


yaronf:
Quoting the article:

> I was young at the time, and the whole [discussion] struck me as opaque, puritanical and alienating, so I pretty much gave up on the whole idea of REST: it was something condescending people fought about on the internet.

I am not as young but I agree with this sentiment. Having said that, I guess a standard needs to be reasonably precise in its usage of terminology, and so I accept the PR.

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -114,7 +114,7 @@ policy management and message authorization are out of scope of this document.
 The high-level message flow is as follows:
 
 * Workload A obtains a credential from the Identity Server. This happens periodically, e.g. once every 24 hours.
-* Workload A makes a REST call into Workload B. This is a regular REST call, with the additional protection
+* Workload A makes an HTTP call into Workload B. This is a regular HTTP call, with the additional protection
```
yaronf, line 117: I don't think I've ever heard the term "HTTP call". Can we think of a better term?

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -56,15 +56,15 @@ TODO Abstract
 
 This document defines authentication and authorization in the context of interaction between two workloads.
 This is the core component of the WIMSE architecture {{?I-D.ietf-wimse-arch}}.
-Assume that Service A needs to call Service B. For simplicity, this document focuses on REST services,
+Assume that Service A needs to call Service B. For simplicity, this document focuses on HTTP services,
```
yaronf, line 59:
```suggestion
Assume that Service A needs to call Service B. For simplicity, this document focuses on HTTP-based services,
```

yaronf (changes requested):

draft-sheffer-wimse-s2s-protocol.md
bc-pi: I feel like I hear the term used all the time but that doesn't necessarily it's widely used or understood. How about just good ol' "HTTP request"? 

```suggestion
* Workload A makes an HTTP call into Workload B. This is a regular HTTP request, with the additional protection
```

PR #9, opened 2024-06-07
Initial version of Message Signatures section

yaronf:
A profile of the Message Signatures RFC. Still missing an example (because I wanted to generate the example using the key material in the ID Token section, and Brian keeps his secret key secret).

### DIFF
--- a/draft-sheffer-wimse-s2s-protocol.md
+++ b/draft-sheffer-wimse-s2s-protocol.md
@@ -136,6 +136,51 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
 
 ## Option 2: Authentication Based on HTTP Message Signatures
 
+This option uses the WIMSE Identity Token (ref TBD) to sign the request and optionally, the response.
+This specification defines a profile of the Message Signatures specification {{!RFC9421}}.
+
+The request is signed as per {{RFC9421}}. The following derived components MUST be signed:
+
+* `@method`
+* `@request-target`
+
+In addition, the following headers MUST be signed when they exist:
+
+* `content-type`
+* `content-digest`
+
+If the response is signed, the following components MUST be signed:
+
+* `@status`
+* `@method;req`
+* `@request-target;req`
+* `content-type` if it exists
+* `content-digest` if it exists
+
+For both requests and responses, the following signature parameters MUST be included:
+
+* `created`
+* `expires` - expiration MUST be short, e.g. on the order of minutes. The WIMSE architecture will provide seperate
+mechanisms in support of long-lived compute processes.
+* `nonce`
+* `alg`
+* `tag` - the value for implementations of this specification is `wimse-service-to-service`
+
+Since the signing key is sent along with the message, the `keyid` parameter SHOULD NOT be used.
+
+It is RECOMMENDED to include only one signature with the HTTP message.
+If multiple ones are included, then the signature label included in both the `Signature-Input` and `Signature` headers SHOULD
+be `wimse`.
+
+To promote interoperability, the `ecdsa-p256-sha256` signing algorithm MUST be implemented
+by general purpose implementations of this spec.
+
+The `Accept-Signature` field is not used by this profile
+
+Following is a non-normative example of a signed request and a signed response, using the keys mentioned in Section TBD.
+
+TODO
+
 # Using Mutual TLS for Service To Service Authentication {#mutual-tls}
 
 # Security Considerations


draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -136,6 +136,65 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+* TBD that includes the WIT
+
+If the response is signed, the following components MUST be signed:
+
+* `@status`
+* `@method;req`
+* `@request-target;req`
+* `content-type` if it exists
+* `content-digest` if it exists
+* `Authorization`
```
bc-pi, line 161: The Authorization header is only applicable to requests, no?  
```suggestion
```

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -136,6 +136,65 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+
+~~~ http
+GET /gimme-ice-cream?flavor=vanilla HTTP/1.1
+Host: example.com
+Authorization: Basic c3BpZmZlOi8vcmVhbG0uZXhhbXBsZS5jb20vc3ZjYmMxMjM=
+Signature: wimse=:cJktyi42FuhfTjuJl5vlwHMkQggviWQD/ho1UKMy96VtOeECRrc \
+UhxlabRRaTgXdRJrzVR7IAfvR7/8E4y7yBg==:
+Signature-Input: wimse=("@method" "@request-target" "authorization"); \
+created=1717799748;expires=1717800048; \
+nonce="abcd1111";tag="wimse-service-to-service";keyid="dummy-keyid-TODO"
```
bc-pi, line 194: above it says that, "the `keyid` parameter SHOULD NOT be used." but there's a keyid in this line of the example

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -136,6 +136,65 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+by general purpose implementations of this spec.
+
+The `Accept-Signature` field is not used by this profile
+
+Following is a non-normative example of a signed request and a signed response, using the keys mentioned in Section TBD.
+
+~~~ http
+GET /gimme-ice-cream?flavor=vanilla HTTP/1.1
+Host: example.com
+Authorization: Basic c3BpZmZlOi8vcmVhbG0uZXhhbXBsZS5jb20vc3ZjYmMxMjM=
```
bc-pi, line 189: c3BpZmZlOi8vcmVhbG0uZXhhbXBsZS5jb20vc3ZjYmMxMjM= decodes to `spiffe://realm.example.com/svcbc123` which doesn't seem right for HTTP basic. To the extent that HTTP basic even seems right in an example like this.

draft-sheffer-wimse-s2s-protocol.md
bc-pi: also says alg is required above but doesn't seem to be in the example

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -136,6 +136,65 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+It is RECOMMENDED to include only one signature with the HTTP message.
+If multiple ones are included, then the signature label included in both the `Signature-Input` and `Signature` headers SHOULD
+be `wimse`.
+
+To promote interoperability, the `ecdsa-p256-sha256` signing algorithm MUST be implemented
+by general purpose implementations of this spec.
+
+The `Accept-Signature` field is not used by this profile
+
+Following is a non-normative example of a signed request and a signed response, using the keys mentioned in Section TBD.
```
bc-pi, line 184: looks like only example of a request ?

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -136,6 +136,65 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+The request is signed as per {{RFC9421}}. The following derived components MUST be signed:
+
+* `@method`
+* `@request-target`
+
+In addition, the following headers MUST be signed when they exist:
+
+* `content-type`
+* `content-digest`
+* `Authorization`
```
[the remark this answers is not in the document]
yaronf: > should we include `Txn-Token`? A fair number of folks seem to think it might be used to convey transational context info https://www.ietf.org/archive/id/draft-ietf-oauth-transaction-tokens-01.html#name-txn-token-http-header

Yes.

draft-sheffer-wimse-s2s-protocol.md
yaronf: > above it says that, "the `keyid` parameter SHOULD NOT be used." but there's a keyid in this line of the example

This was a problem with my own implementation of Message Sigs, now fixed.

yaronf: > also says `alg` is required above but doesn't seem to be in the example

In fact I think that `alg` as a Message Sig parameter is a mistake and a vulnerability waiting to happen (because verifiers would trust this field even though they lack context, similarly to the "none" issue with JWTs. IMO an algorithm should be strongly bound to a key. Can we make the `alg` field mandatory within the JWK?

draft-sheffer-wimse-s2s-protocol.md
yaronf: > c3BpZmZlOi8vcmVhbG0uZXhhbXBsZS5jb20vc3ZjYmMxMjM= decodes to `spiffe://realm.example.com/svcbc123` which doesn't seem right for HTTP basic. To the extent that HTTP basic even seems right in an example like this.

@bc-pi I need your help with this one. (And I thought that the string also includes a "password").

draft-sheffer-wimse-s2s-protocol.md
yaronf: Added the response, signed by the responder of course.

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -136,6 +136,89 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+* `@method`
+* `@request-target`
+
+In addition, the following headers MUST be signed when they exist:
+
+* `Content-Type`
+* `Content-Digest`
+* `Authorization`
+* `Txn-Token` {{?I-D.ietf-oauth-transaction-tokens}}
+* TBD that includes the WIT
```
jsalowey, line 153: Do we envision cases where the WIT would not be included?

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -136,6 +136,89 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+Following is a non-normative example of a signed request and a signed response, using the keys mentioned in Section TBD.
+
+~~~ http
+GET /gimme-ice-cream?flavor=vanilla HTTP/1.1
+Host: example.com
+Authorization: Basic c3BpZmZlOi8vcmVhbG0uZXhhbXBsZS5jb20vc3ZjMTphYmMxMjM=
+Signature: wimse=:L0xn/2/XncJ0QYNYuwvPsDGkJ6Gbe+rFA9soJlDfXcsqEfC7enVXFIHqCGJM7gtG6kukZUw0j/YaSXmDOiaZCQ==:
+Signature-Input: wimse=("@method" "@request-target" "authorization");created=1718102553;expires=1718102853;nonce="abcd1111";tag="wimse-service-to-service"
+
+~~~
```
jsalowey, line 196: Shouldn't the example include the WIT?

draft-sheffer-wimse-s2s-protocol.md
yaronf: @jsalowey No I don't, just waiting for the header definition from @bc-pi.

draft-sheffer-wimse-s2s-protocol.md
yaronf: @jsalowey As above: yes when the header is defined.

draft-sheffer-wimse-s2s-protocol.md
yaronf: New issue: #15

draft-sheffer-wimse-s2s-protocol.md
yaronf: `keyid` comment addressed. `alg` comment now #16.

PR #11, opened 2024-06-09
initial security considerations

jsalowey:


draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -140,9 +140,50 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
 
 # Security Considerations
 
-TODO Security and Privacy
+## WIMSE ID
```
yaronf, line 143: I think "WIMSE Identity" is clearer.

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -140,9 +140,50 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+
+The WIMSE ID is scoped within a trust domain and therefore a WIMSE ID is only unique within a trust domain. Using a WIMSE ID without taking into account the trust domain could allow one domain to issue tokens to spoof identities in another domain.  Additionally, the trust domain must be tied to an authorized trusted key through some mechanism such as a JWKS or X.509 certificate chain. The association of a trust domain and a cryptographic trust root MUST be communicated securely out of band.
```
yaronf, line 145: The first two sentences are confusing because the (proposed) WIMSE Identity *includes* the trust domain.

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -140,9 +140,50 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+
+[TODO: Should there be a DNS name to Trust domain mapping defined?]
+
+## WIMSE ID Token and Proof of Possession
+
+The WIMSE ID token is bound to a secret cryptographic key and is always presented with a proof of possession as described in section TBD. It MUST NOT be used as a bearer token. While this helps reduce the sensitivity of the token it is still possible that a token and its proof of possession may be captured and replayed.  If this risk is mitigated then WIMSE ID tokens may be resued for a period of time with fresh proofs generated for each transaction.  The following are some mitigations for the capture and reuse of the proof of possession (POP):
```
yaronf, line 151: Make the first sentence conditional: the ID Token and its PoP are only used in the application-level options, and both are not used in MTLS.

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -140,9 +140,50 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+
+* Preventing Eavesdropping and Interception with TLS
+
+An attacker observing or intercepting the communication channel can view the token and its proof of possession and attempt to replay it to gain an advantage. In order to prevent this the
+token and proof of possession MUST be sent over a secure, server authenticated TLS connection unless a secure channel is provided by some other mechanisms. Host name validation according
+to section TBD MUST be performed. The WIMSE ID token itself is not usable without a proof of possession.
```
yaronf, line 157: Hmmm, not exactly. Sometimes there's a middlebox rather than end-to-end TLS, and then you have host name validation, but the "host" is the middlebox.

yaronf: Also, the Message Sig section includes replay mitigations (detection of duplicate nonces). We should mention that and have a similar mechanism in the pop-esque option.

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -140,9 +140,50 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+
+* Limiting Proof of Possession Lifespan
+
+The proof of possession MUST be time limited. A POPs should only for the time necessary for it to be successfully for the purpose it is needed. This will typically be on the order of minutes.  POPs received outside their validity time MUST be rejected.
```
yaronf, line 161:
```suggestion
The proof of possession MUST be time limited. A POP should only be valid over the time necessary for it to be successfully used for the purpose it is needed. This will typically be on the order of minutes.  POPs received outside their validity time MUST be rejected.
```

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -140,9 +140,50 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+
+* Limiting Proof of Possession Scope
+
+The POP MUST have a limited scope. Limitations on scope include constraining the POP to certain receivers and constraining the POP to a usage or transaction.
```
yaronf, line 165: This is true of the message, I'm not sure it is true of the POP.

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -140,9 +140,50 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+
+* Binding to a Nonce
+
+The POP MAY be bound to a fresh nonce to provide single use properties.  The mechanism for distributing a nonce are outside the scope of this specification.
```
yaronf, line 169: A nonce is useful even when it is not "distributed". Even a nonce generated by the message originator can be used for limited replay protection.

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -140,9 +140,50 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+
+## Privacy Considerations
+
+WIMSE tokens and the proofs of possession may contain private information such as user names or other identities. Care should be taken to prevent the disclosure of this information. The use of TLS helps protect the privacy of WIMSE tokens and proofs of possession.
+
+WIMSE tokens are typically associated with a workload that is not directly associated with a user, however in some deployments the workload may be associated directly to a user. In these cases the WIMSE token can be used to track the user if it is disclosed in clear text.
+
+If the WIMSE ID is used with X509 certificates then the workload identity may be disclosed as part of the TLS handshake.  This can be partially mitigated by using TLS 1.3 to protect the client and serve identities.
```
yaronf, line 185:
```suggestion
If the WIMSE ID is used with X.509 certificates then the workload identity may be disclosed as part of the TLS handshake.  This can be partially mitigated by using TLS 1.3 to protect the client and server identities.
```

yaronf: I don't think that protection of workload identity falls under the "privacy" title. In the service-to-service case, user identity may exist as "context", and this is mentioned above. In the user-to-service case, this would typically be OAuth rather than WIMSE.

yaronf (changes requested):

draft-sheffer-wimse-s2s-protocol.md
jsalowey: You still need to do host name validation.  the service will bee to know the name of the intermediary.  I'm not sure there is more to explain here. 
I added a bit about the nonces in a later section.

draft-sheffer-wimse-s2s-protocol.md
jsalowey: I think this is probably something we should discuss.  There is a security trade-off broadening the scope of the POP.  I think we can evolve this section to discuss the trade-offs

draft-sheffer-wimse-s2s-protocol.md
jsalowey: The connection I made to privacy is that it may be possible that a workload directly maps to a user in some cases.  In this case the workload identity can be used to track the behavior of a user.  To me this is a corner case, but I thought it was worth mentioning.

arndt-s (approved):

draft-sheffer-wimse-s2s-protocol.md
jsalowey: Clarified this applies to the subcomponent of a name are only unique within a trust domain.

draft-sheffer-wimse-s2s-protocol.md
jsalowey: applied

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -140,9 +140,49 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+
+## Privacy Considerations
+
+WITs and the proofs of possession may contain private information such as user names or other identities. Care should be taken to prevent the disclosure of this information. The use of TLS helps protect the privacy of WITs and proofs of possession.
+
+WITs and Certificates with WIMSE identifiers are typically associated with a workload and not a specific user, however in some deployments the workload may be associated directly to a user. While these are exceptional cases a deployment should evaluate if the disclosure of WITs or certificates can be used to track a user.
```
yaronf, line 184:
```suggestion
WITs and certificates with WIMSE identifiers are typically associated with a workload and not a specific user, however in some deployments the workload may be associated directly to a user. While these are exceptional cases a deployment should evaluate if the disclosure of WITs or certificates can be used to track a user.
```

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -140,9 +140,49 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+
+* Binding to a Timestamp or Nonce
+
+A proof of possession should include information that can be used to uniquely identify it such as a unique timestamp or nonce.  This can be used by the receiver to perform basic replay protection against tokens it has already seen. Depending upon the design of the system it may be difficult to synchronize the replay cache across all token validators. In this case, if the POP is not sufficiently scoped it may be usable with another workload.
+While a fresh nonce could be included in the POP, a mechanism for distributing a fresh challenge nonce from the validator to provide single use properties of a POP is outside the scope of this specification.
```
yaronf, line 170:
```suggestion
The proof of possession MUST be time limited. A PoP should only be valid over the time necessary for it to be successfully used for the purpose it is needed. This will typically be on the order of minutes.  PoPs received outside their validity time MUST be rejected.

* Limiting Proof of Possession Scope

In order to reduce the risk of theft and replay the PoP should have a limited scope. For example, a PoP may be targeted for use with a specific workload and even a specific transaction to reduce the impact of a stolen PoP. In some cases a workload may wish to reuse a PoP for a period of time or have it accepted by multiple target workloads. A careful analysis is warranted to understand the impacts to the system if a PoP is disclosed allowing it to be presented by an attacker along with a captured WIT.

* Binding to a Timestamp or Nonce

A proof of possession should include information that can be used to uniquely identify it such as a unique timestamp or nonce.  This can be used by the receiver to perform basic replay protection against tokens it has already seen. Depending upon the design of the system it may be difficult to synchronize the replay cache across all token validators. In this case, if the PoP is not sufficiently scoped it may be usable with another workload.
While a fresh nonce could be included in the PoP, a mechanism for distributing a fresh challenge nonce from the validator to provide single use properties of a PoP is outside the scope of this specification.
```

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -140,9 +140,49 @@ This document uses "service" and "workload" interchangeably. Otherwise, all term
...
+
+* Binding to Sender
+
+The POP MAY be bound to a sender such as the client identity of a TLS session or TLS channel binding parameters. The mechanisms for binding are outside the scope of this specification.
+
+## Middle Boxes
+
+In some deployments the WIMSE token and proof of possession may pass through multiple systems. The communication between the systems is over TLS, but the token and POP are available in the clear at each intermediary.  While the intermediary cannot modify the token or the information within the POP they can attempt to capture and replay the token or modify the data not protected by the POP. Mitigations listed in the previous section can be used to provide some protection from middle boxes. Deployments should perform analysis on their situation to determine if it is appropriate to trust and allow traffic to pass through a middle box.
```
yaronf, line 178:
```suggestion
In some deployments the WIMSE token and proof of possession may pass through multiple systems. The communication between the systems is over TLS, but the token and PoP are available in the clear at each intermediary.  While the intermediary cannot modify the token or the information within the PoP they can attempt to capture and replay the token or modify the data not protected by the PoP. Mitigations listed in the previous section can be used to provide some protection from middle boxes. Deployments should perform analysis on their situation to determine if it is appropriate to trust and allow traffic to pass through a middle box.
```

yaronf (changes requested):

PR #12, opened 2024-06-10
Service->Workload

frumioj:
I do not think we should use "workload" and "service" interchangeably. They MAY be synonymous in some deployments. But in some cases, elements of a workload will be heterogeneous and despite having the same "identity" (as a workload) will look quite different from an external perspective (e.g. the API).


yaronf (commented):
Thank you @frumioj. We discussed this very issue within the team and decided to put it on hold until after -00 is published (a.k.a., naming things is hard). But feel free to tag others and get their view.

arndt-s:
Closing this for https://github.com/yaronf/wimse-s2s/issues/13 - if we decide to rename a new PR needs to be provided on top of the upcoming commits from the other PRs anyway.

PR #19, opened 2024-06-19
First cut at the DPoP-Inspired Authentication section

bc-pi:
> Yaron Sheffer
  [8:31 AM May 30 in slack]
To summarize the near term plan:
Joe: security considerations and interaction with TLS
Brian: ID Token and DPoP-inspired
Yaron: Message Signatures

this is the "DPoP-inspired" part 

A preview editors' copy of this PR can be seen at http://www.sheffer.org/wimse-s2s/bc-dpop-esque-pop/draft-sheffer-wimse-s2s-protocol.html

### DIFF
--- a/draft-sheffer-wimse-s2s-protocol.md
+++ b/draft-sheffer-wimse-s2s-protocol.md
@@ -293,7 +293,7 @@ A WPT contains the following:
 * in the JWT claims:
     * `iss`: The issuer of the token, which is the calling workload, represented by the same value as the `sub` claim
      of the associated WIT.
-    * `aud`: The audience of the token contains the the HTTP target URI ({{Section 7.1 of RFC9110}}) of the request
+    * `aud`: The audience of the token contains the HTTP target URI ({{Section 7.1 of RFC9110}}) of the request
      to which the WPT is attached, without query or fragment parts.
     * `exp`: The expiration time of the WIT (as defined in {{Section 4.1.4 of RFC7519}}). WPT lifetimes MUST be short,
      e.g., on the order of minutes or seconds.
--- a/draft-sheffer-wimse-s2s-protocol.md
+++ b/draft-sheffer-wimse-s2s-protocol.md
@@ -295,7 +295,8 @@ A WPT contains the following:
      of the associated WIT.
     * `aud`: The audience of the token contains the the HTTP target URI ({{Section 7.1 of RFC9110}}) of the request
      to which the WPT is attached, without query or fragment parts.
-    * `iat`: The time at which the token was created (as defined in {{Section 4.1.6 of RFC7519}}).
+    * `exp`: The expiration time of the WIT (as defined in {{Section 4.1.4 of RFC7519}}). WPT lifetimes MUST be short,
+     e.g., on the order of minutes or seconds.
     * `jti`: A unique identifier for the token.
     * `ctx_token_hash`: Hash of the OAuth access token, Transaction Token {{?I-D.ietf-oauth-transaction-tokens}},
      or other token in the request that might convey end-user identity and authorization context of the request.
@@ -307,11 +308,10 @@ An example WPT might look like the following:
 ~~~ jwt
 eyJ0eXAiOiJ3aW1zZS1wcm9vZitqd3QiLCJhbGciOiJFZERTQSJ9.eyJpc3MiOiJ3aW1z
 ZTovL2V4YW1wbGUuY29tL3NwZWNpZmljLXdvcmtsb2FkIiwiYXVkIjoiaHR0cHM6Ly9zZ
-XJ2aWNlLmV4YW1wbGUuY29tL3BhdGgiLCJpYXQiOjE3MTc2MTI1NDUsImp0aSI6Il9fYn
+XJ2aWNlLmV4YW1wbGUuY29tL3BhdGgiLCJleHAiOjE3MTc2MTI4MjAsImp0aSI6Il9fYn
 djNEVTQzNhY2MyTFRDMS1feCIsImN0eF90b2tlbl9oYXNoIjoiQ0w0d2pmcFJtTmYtYmR
-ZSWJZTG5WOWQ1ck1BUkd3S1lFMTB3VXd6QzBqSSJ9.YVS0gGOnsCK5xWhE9lcRzp0CDpl
-TGLDv3lUK4dSF_o2uw2wh6I2FOwsa0OmUMeTAr-qO0bJJBjSOy64UKBekCg
-
+ZSWJZTG5WOWQ1ck1BUkd3S1lFMTB3VXd6QzBqSSJ9.6BFrZdobweg5cg1NHvSKqEtmqka
+3zu78d2E91VKMoO0tP3J0Pwja7GoAOKuU2ztPAH3yfogFyOBS2GKqo4RSDA
 ~~~
 {: #example-wpt title="Example Workload Proof Token (WPT)"}
 
@@ -331,7 +331,7 @@ The decoded JWT claims of the WPT from the example above are shown here:
 {
  "iss": "wimse://example.com/specific-workload",
  "aud": "https://service.example.com/path",
- "iat": 1717612545,
+ "exp": 1717612820,
  "jti": "__bwc4ESC3acc2LTC1-_x",
  "ctx_token_hash": "CL4wjfpRmNf-bdYIbYLnV9d5rMARGwKYE10wUwzC0jI"
 }
@@ -374,7 +374,8 @@ To validate the WPT in the request, the recipient MUST ensure the following:
 * The `iss` claim of the WPT matches the `sub` claim of the WIT. (note: not sure `iss` in the WPT is necessary)
 * The `aud` claim of the WPT matches the target URI, or an acceptable alias or normalization thereof, of the HTTP request
  in which the WPT was received, ignoring any query and fragment parts.
-* The `iat` claim of the WPT is within an acceptable window, e.g. within a few seconds or minutes of the current time.
+* The `exp` claim is present and conveys a time that has not passed. WPTs with an expiration time unreasonably
+ far in the future SHOULD be rejected.
 * Optionally, check that the value of the `jti` claim has not been used before in the time window in which the
  respective WPT would be considered valid.
 * If presented in conjunction with a token conveying end-user identity or authorization context, the value of
--- a/draft-sheffer-wimse-s2s-protocol.md
+++ b/draft-sheffer-wimse-s2s-protocol.md
@@ -298,10 +298,17 @@ A WPT contains the following:
     * `exp`: The expiration time of the WIT (as defined in {{Section 4.1.4 of RFC7519}}). WPT lifetimes MUST be short,
      e.g., on the order of minutes or seconds.
     * `jti`: A unique identifier for the token.
-    * `ctx_token_hash`: Hash of the OAuth access token, Transaction Token {{?I-D.ietf-oauth-transaction-tokens}},
-     or other token in the request that might convey end-user identity and authorization context of the request.
-     The value MUST be the result of a base64url encoding (as defined in {{Section 2 of RFC7515}}) the SHA-256 hash of
+    * `ath`: Hash of the OAuth access token, if present in the request, which might convey end-user identity and
+     authorization context of the request. The value, as per {{Section 4.1 of RFC9449}},
+     is the base64url encoding of the SHA-256 hash of the ASCII encoding of the access token's value.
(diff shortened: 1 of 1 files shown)


draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -279,6 +279,109 @@ etc., are all valid and equivalent header field names. However, case is signific
...
+     of the associated WIT.
+    * `aud`: The audience of the token contains the the HTTP target URI ({{Section 7.1 of RFC9110}}) of the request
+     to which the WPT is attached, without query or fragment parts.
+    * `iat`: The time at which the token was created (as defined in {{Section 4.1.6 of RFC7519}}).
+    * `jti`: A unique identifier for the token.
+    * `ctx_token_hash`: Hash of the OAuth access token, Transaction Token {{?I-D.ietf-oauth-transaction-tokens}},
+     or other token in the request that might convey end-user identity and authorization context of the request.
+     The value MUST be the result of a base64url encoding (as defined in {{Section 2 of RFC7515}}) the SHA-256 hash of
+     the ASCII encoding of the associated token's value.
+
```
jsalowey, line 304: Can there be more than one context token hash?  would this need to be specified as an array or can you have multiple of the same claim in the jwt?

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -279,6 +279,109 @@ etc., are all valid and equivalent header field names. However, case is signific
...
+A WPT contains the following:
+
+* in the JOSE header:
+    * `alg`: An identifier for an appropriate JWS asymmetric digital signature algorithm corresponding to
+     the confirmation key in the associated WIT.
+    * `typ`: the WPT is explicitly typed, as recommended in {{Section 3.11 of RFC8725}},
+     using the `application/wimse-proof+jwt` media type.
+* in the JWT claims:
+    * `iss`: The issuer of the token, which is the calling workload, represented by the same value as the `sub` claim
+     of the associated WIT.
```
jsalowey, line 295: You might want to also include the hash of the WIT in the proof of possession.  Perhaps just having the full name is good enough, but I think you want to account for the issuer of the WIT as well in the signature.

draft-sheffer-wimse-s2s-protocol.md
bc-pi: > Can there be more than one context token hash? 

This is a good question that I asked myself while writing the above. But didn't have a great sense for so opted for the simple answer of just one. 

The whole construct is admittedly somewhat awkward - trying to account/allow for different kinds of tokens that might show up in different places. 

 
> would this need to be specified as an array or can you have multiple of the same claim in the jwt?

Multiple of the same claim in the jwt is not allowed (see https://datatracker.ietf.org/doc/html/rfc7519#section-4) so an array is how that kind of data is expressed.

bc-pi: But maybe a few different claims that conditionally show up would be a better approach. Like `ath` access token hash - a la https://www.rfc-editor.org/rfc/rfc9449.html#section-4.2-6.2 - and then `txth` transaction token hash and maybe an oth for *other* token hash...

draft-sheffer-wimse-s2s-protocol.md
bc-pi: > You might want to also include the hash of the WIT in the proof of possession. 

What purpose would that serve? Asking for a friend. 

> Perhaps just having the full name is good enough, but I think you want to account for the issuer of the WIT as well in the signature.

I don't think I follow the thinking here...

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -279,6 +279,109 @@ etc., are all valid and equivalent header field names. However, case is signific
...
+    * `aud`: The audience of the token contains the the HTTP target URI ({{Section 7.1 of RFC9110}}) of the request
+     to which the WPT is attached, without query or fragment parts.
+    * `iat`: The time at which the token was created (as defined in {{Section 4.1.6 of RFC7519}}).
```
bc-pi, line 298: Consider exp rather than iat

draft-sheffer-wimse-s2s-protocol.md
bc-pi: action from the call was to put in an issue for further discussion and proceed without a WIT hash for now

draft-sheffer-wimse-s2s-protocol.md
bc-pi: Currently rethinking the construct

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -279,6 +279,109 @@ etc., are all valid and equivalent header field names. However, case is signific
...
+* in the JOSE header:
+    * `alg`: An identifier for an appropriate JWS asymmetric digital signature algorithm corresponding to
+     the confirmation key in the associated WIT.
+    * `typ`: the WPT is explicitly typed, as recommended in {{Section 3.11 of RFC8725}},
+     using the `application/wimse-proof+jwt` media type.
+* in the JWT claims:
+    * `iss`: The issuer of the token, which is the calling workload, represented by the same value as the `sub` claim
+     of the associated WIT.
+    * `aud`: The audience of the token contains the the HTTP target URI ({{Section 7.1 of RFC9110}}) of the request
+     to which the WPT is attached, without query or fragment parts.
```
yaronf, line 297: Why isn't `aud` the WIMSE identity of the target? Wouldn't that be more consistent?

draft-sheffer-wimse-s2s-protocol.md
yaronf: Although we can definitely punt on it for now, I think we could choose a construct that's both simple and extensible.

```json
"signed_headers": {
    "txn-token": "txn token hash",
    "x-custom-context": "x-custom-context hash"
}
```

Unfortunately we need the WPT to sign miscellaneous stuff, because we don't want a new WIT for every call. Having said that, as far as I can tell it does NOT need to sign the WIT itself.

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -279,6 +279,109 @@ etc., are all valid and equivalent header field names. However, case is signific
...
+ VS0gGOnsCK5xWhE9lcRzp0CDplTGLDv3lUK4dSF_o2uw2wh6I2FOwsa0OmUMeTAr-qO0
+ bJJBjSOy64UKBekCg
+
+{"do stuff":"please"}
+~~~
+{: title="Example HTTP Request with WIT and WPT"}
+
+To validate the WPT in the request, the recipient MUST ensure the following:
+
+* There is not more than one `Workload-Proof-Token` header field in the request.
```
yaronf, line 370:
```suggestion
* There is exactly one `Workload-Proof-Token` header field in the request.
```

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -279,6 +279,109 @@ etc., are all valid and equivalent header field names. However, case is signific
...
+* The `Workload-Proof-Token` header field value is a single and well-formed JWT.
+* The WPT signature is valid using the public key from the confirmation claim of the WIT.
+* The `typ` JOSE header parameter of the WPT conveys a media type of `wimse-proof+jwt`.
+* The `iss` claim of the WPT matches the `sub` claim of the WIT. (note: not sure `iss` in the WPT is necessary)
+* The `aud` claim of the WPT matches the target URI, or an acceptable alias or normalization thereof, of the HTTP request
+ in which the WPT was received, ignoring any query and fragment parts.
```
yaronf, line 376: Man, what's a poor implementer to do with "an acceptable alias or normalization thereof"?

yaronf (changes requested):

draft-sheffer-wimse-s2s-protocol.md
bc-pi: That, to me, begs a larger philosophical and scoping question about naming and addressing and identity, which is beyond my competence. And in that lacking of competence, I've used an existing, well-known, and widely used means of identifying the target of an HTTP request.

draft-sheffer-wimse-s2s-protocol.md
bc-pi: issue here https://github.com/yaronf/wimse-s2s/issues/24 for posterity and resolving this conversation

draft-sheffer-wimse-s2s-protocol.md
bc-pi: done w/ 26dff3d954a018216c69f66dce18274ee7f56c1d

draft-sheffer-wimse-s2s-protocol.md
yaronf: Let's keep it for now and I will open an issue.

bc-pi: #26 for the folks playing along at home

draft-sheffer-wimse-s2s-protocol.md
yaronf: Hmmph, let's discuss offline. In the meantime, please remove the redundant "the" from the sentence.

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -279,6 +279,110 @@ etc., are all valid and equivalent header field names. However, case is signific
...
+
+* in the JOSE header:
+    * `alg`: An identifier for an appropriate JWS asymmetric digital signature algorithm corresponding to
+     the confirmation key in the associated WIT.
+    * `typ`: the WPT is explicitly typed, as recommended in {{Section 3.11 of RFC8725}},
+     using the `application/wimse-proof+jwt` media type.
+* in the JWT claims:
+    * `iss`: The issuer of the token, which is the calling workload, represented by the same value as the `sub` claim
+     of the associated WIT.
+    * `aud`: The audience of the token contains the the HTTP target URI ({{Section 7.1 of RFC9110}}) of the request
```
bc-pi, line 296: pizza pizza
```suggestion
    * `aud`: The audience of the token contains the HTTP target URI ({{Section 7.1 of RFC9110}}) of the request
```

draft-sheffer-wimse-s2s-protocol.md
bc-pi: done 06948498e55d9cdf8eccc34fc00c2c7c135f2cc0 and looking forward to a lively discussion

draft-sheffer-wimse-s2s-protocol.md
bc-pi: https://github.com/yaronf/wimse-s2s/issues/25

bc-pi: Created issue #25 from @yaronf's comment and c5b46a1d06829b5ec435f143d16d57a478003706 tries to do this in a somewhat better way.

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -279,6 +279,119 @@ etc., are all valid and equivalent header field names. However, case is signific
...
+     to which the WPT is attached, without query or fragment parts.
+    * `exp`: The expiration time of the WIT (as defined in {{Section 4.1.4 of RFC7519}}). WPT lifetimes MUST be short,
+     e.g., on the order of minutes or seconds.
+    * `jti`: A unique identifier for the token.
+    * `ath`: Hash of the OAuth access token, if present in the request, which might convey end-user identity and
+     authorization context of the request. The value, as per {{Section 4.1 of RFC9449}},
+     is the base64url encoding of the SHA-256 hash of the ASCII encoding of the access token's value.
+    * `tth`: Hash of the Txn-Token {{?I-D.ietf-oauth-transaction-tokens}}, if present in the request,
+     which might convey end-user identity and authorization context of the request. The value MUST be the result of
+     a base64url encoding (as defined in {{Section 2 of RFC7515}}) the SHA-256 hash of
```
yaronf, line 306:
```suggestion
     a base64url encoding (as defined in {{Section 2 of RFC7515}}) of the SHA-256 hash of
```

draft-sheffer-wimse-s2s-protocol.md
```diff
@@ -279,6 +279,119 @@ etc., are all valid and equivalent header field names. However, case is signific
...
+     the ASCII encoding of the associated token's value.
+    * `oth`: Hash of any other token in the request that might convey end-user identity and authorization context of the
+     request. The value MUST be the result of a base64url encoding (as defined in {{Section 2 of RFC7515}}) the
```
yaronf, line 309:
```suggestion
     request. The value MUST be the result of a base64url encoding (as defined in {{Section 2 of RFC7515}}) of the
```

yaronf (approved):

PR #33, opened 2024-07-03
Fold long lines

yaronf:
So we don't fail id-nits.


yaronf:
The result is simply ugly. Closing and replacing by rfcfold.

PR #55, opened 2024-07-24
Change trust root to trust anchor in draft-sheffer-wimse-s2s-protocol.md

kenmccracken-google:
Aligns with Trust Anchor terminology as used in [rfc5280](https://datatracker.ietf.org/doc/html/rfc5280).

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,4 @@
 # License
 
 See the
-[guidelines for contributions](https://github.com/yaronf/wimse-s2s/blob/main/CONTRIBUTING.md).
+[guidelines for contributions](https://github.com/ietf-wg-wimse/draft-ietf-wimse-s2s-protocol/blob/main/CONTRIBUTING.md).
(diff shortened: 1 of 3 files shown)


yaronf:
I would defer to @jsalowey here, who wrote the original text and I suspect is also more of an expert on this terminology.

kenmccracken-google:
This seems to have bitrot.  Please refer to https://github.com/ietf-wg-wimse/draft-ietf-wimse-s2s-protocol/pull/58 instead.

yaronf:
@kenmccracken-google Hi Ken, why did you force-push into the "main" branch?

bc-pi:
> @kenmccracken-google Hi Ken, why did you force-push into the "main" branch?

is that what happened here? I don't see where you see that that happened?

yaronf:
@bc-pi I see it on the GH UI right above here, but apparently it failed because the commit log doesn't show it.

kenmccracken-google:
I apologize, the force-push was an error, I probably clicked the wrong options.  AFAICT the outcome is not in the commit history.

Issue #24, opened 2024-06-26
might want to also include the hash of the WIT in the proof


bc-pi:
>  You might want to also include the hash of the WIT in the proof of possession.  Perhaps just having the full name is good enough, but I think you want to account for the issuer of the WIT as well in the signature.

_Originally posted by @jsalowey in https://github.com/yaronf/wimse-s2s/pull/19#discussion_r1646749641_

jsalowey:
If the WIT is not included in the signature I think you can end up in a situation where the client and server have a different understanding of what was authenticated on the connection.  The workload identity is arguably the most important piece of information which is already included in the signature, but the WIT contains other information as well. For example the issuer that signed that token although this should be mapped to the workload ideneity there could be differences in some cases.  In addition if the WIT contains any other fields that contain important information then the client and server can have processed.  
I'm pretty convinced that we should include the WIT in the signature.  Perhaps we can get away with it if the WIT really only contains the workload identity and not other information (no extensions) and we are confident that deployments will get the trust root configuration for token validation correct.

yaronf:
Maybe we should clarify that a WIT must only contain identity information, and not authorization/policy. But we all know that people ignore such warnings.

PieterKas:
@joe-venafi is there an attack you can describe that may result from not signing the WIT (even if it does not include other claims). For example, is it possible to swap out a proof from one domain to another if the issuer information is not taken into account (would require the key material to be duplicated)? If there is, we should note it in the security considerations as motivation for including the WIT hash.

arndt-s:
@PieterKas I can give this a try:

Situation:
```mermaid
flowchart LR
WorkloadA --> ReverseProxy --> WorkloadB
```

1. WorkloadA -> ReverseProxy
   ```
   Workload-Identity-Token: T1
   Workload-Proof-Token: P1
   ```

2. ReverseProxy -> WorkloadB
ReverseProxy swaps T1 (or P1) to T2 (or P2).
   ```
   Workload-Identity-Token: T2
   Workload-Proof-Token: P1
   ```

The attack is not as simple as this, T2 claims still need to match P1, but looking at the theory:
**there is no guarantee that the P1 was actually generated for T1 and not T2. As long as claims of the WIT match the proof it is accepted.**

jsalowey:
[...]

PR #57, opened 2024-08-25
Added security considerations for including WIT in signature

jsalowey:
Added security considerations for binding WIT to POP signature. Related to issue #24

### DIFF
--- a/draft-ietf-wimse-s2s-protocol.md
+++ b/draft-ietf-wimse-s2s-protocol.md
@@ -542,7 +542,7 @@ If the system ensures that the information associated with a WIT is the same for
 
 If the proof of possession covers the entire WIT in the signature then the signature will fail if the WIT is substituted and the attacker will not be able to replace the WIT.
 
-The last option to sign the WIT as part of the POP requires the least special verification steps. This is the approach taken by http message signatures in {{http-sig-auth}}. THe DPOP mechanism should take a similar approach if it continues to be part of the proposal.
+The last option to sign the WIT as part of the POP requires the fewest special verification steps. This is the approach taken by http message signatures in {{http-sig-auth}}. The DPOP mechanism should take a similar approach if it continues to be part of the proposal.
 
 * Binding to a Timestamp or Nonce
 
--- a/draft-ietf-wimse-s2s-protocol.md
+++ b/draft-ietf-wimse-s2s-protocol.md
@@ -534,7 +534,7 @@ In order to reduce the risk of theft and replay the PoP should have a limited sc
 
 * Binding WIT to Proof of Possession
 
-If the WIT itself is not bound in the proof of possession then it is possible that the WIT could be replaced by an attacker. If multiple WITs with different claims is issued for the same key this could result in the proof of possession having different meaning than what the sender intended and could give the attacker an advantage. Here are some possible mitigations.
+If the WIT itself is not bound in the proof of possession then it is possible that the WIT could be replaced by an attacker. If multiple WITs with different claims are issued for the same key this could result in the proof of possession having different meaning than what the sender intended and could give the attacker an advantage. Here are some possible mitigations.
 
 If the WIT only contains information that is already included in the POP signature then the different in WITs would not be meaningful. The verifier must check that the signed information matches the information in the WIT. Care must also be taken to understand what is substantial information. For example, even a unique identifier such as a jti claim could be problematic if the system associates additional information to this identifier.
 
--- a/draft-ietf-wimse-s2s-protocol.md
+++ b/draft-ietf-wimse-s2s-protocol.md
@@ -540,7 +540,7 @@ If the WIT only contains information that is already included in the POP signatu
 
 If the system ensures that the information associated with a WIT is the same for a given key that is embedded in the WIT then the attacker will not be able to substitute a different WIT with the signature. Here to care must be taken to understand what is substantial difference is.
 
-If the proof of possession includes the entire WIT in the signature then the signature will fail if the WIT is substituted and the attacker will not be able to replace the WIT.
+If the proof of possession covers the entire WIT in the signature then the signature will fail if the WIT is substituted and the attacker will not be able to replace the WIT.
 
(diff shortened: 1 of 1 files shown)


yaronf:
IMO this text is better placed in an issue rather than a PR, because it calls for discussion between the editors (or on the mailing list). In particular the last paragraph makes it very clear.

So, I suggest to close the PR and move the text back to #24.

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -532,6 +532,18 @@ The proof of possession MUST be time limited. A PoP should only be valid over th
 
 In order to reduce the risk of theft and replay the PoP should have a limited scope. For example, a PoP may be targeted for use with a specific workload and even a specific transaction to reduce the impact of a stolen PoP. In some cases a workload may wish to reuse a PoP for a period of time or have it accepted by multiple target workloads. A careful analysis is warranted to understand the impacts to the system if a PoP is disclosed allowing it to be presented by an attacker along with a captured WIT.
 
+* Binding WIT to Proof of Possession
+
+If the WIT itself is not bound in the proof of possession then it is possible that the WIT could be replaced by an attacker. If multiple WITs with different claims is issued for the same key this could result in the proof of possession having different meaning than what the sender intended and could give the attacker an advantage. Here are some possible mitigations.
```
bc-pi, line 537:
```suggestion
If the WIT itself is not bound in the proof of possession then it is possible that the WIT could be replaced by an attacker. If multiple WITs with different claims are issued for the same key this could result in the proof of possession having different meaning than what the sender intended and could give the attacker an advantage. Here are some possible mitigations.
```

bc-pi, line 537: I feel like a little more qualification around the prerequisites or capabilities of such an attacker is needed here.

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -532,6 +532,18 @@ The proof of possession MUST be time limited. A PoP should only be valid over th
...
+
+If the WIT only contains information that is already included in the POP signature then the different in WITs would not be meaningful. The verifier must check that the signed information matches the information in the WIT. Care must also be taken to understand what is substantial information. For example, even a unique identifier such as a jti claim could be problematic if the system associates additional information to this identifier.
+
+If the system ensures that the information associated with a WIT is the same for a given key that is embedded in the WIT then the attacker will not be able to substitute a different WIT with the signature. Here to care must be taken to understand what is substantial difference is.
+
+If the proof of possession includes the entire WIT in the signature then the signature will fail if the WIT is substituted and the attacker will not be able to replace the WIT.
```
bc-pi, line 543: i think mabye?
```suggestion
If the proof of possession covers the entire WIT in the signature then the signature will fail if the WIT is substituted and the attacker will not be able to replace the WIT.
```

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -532,6 +532,18 @@ The proof of possession MUST be time limited. A PoP should only be valid over th
...
+
+The last option to sign the WIT as part of the POP requires the least special verification steps. This is the approach taken by http message signatures in {{http-sig-auth}}. THe DPOP mechanism should take a similar approach if it continues to be part of the proposal.
```
bc-pi, line 545: "least special" reads awkwardly to me - do you mean like "fewest special"? 

Also I was about to ask "isn't this the same as the prior option?" but then I realized I think that you are talking about the prior option here rather than describing a new option.

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -532,6 +532,18 @@ The proof of possession MUST be time limited. A PoP should only be valid over th
...
+
+If the WIT only contains information that is already included in the POP signature then the different in WITs would not be meaningful. The verifier must check that the signed information matches the information in the WIT. Care must also be taken to understand what is substantial information. For example, even a unique identifier such as a jti claim could be problematic if the system associates additional information to this identifier.
+
+If the system ensures that the information associated with a WIT is the same for a given key that is embedded in the WIT then the attacker will not be able to substitute a different WIT with the signature. Here to care must be taken to understand what is substantial difference is.
+
+If the proof of possession covers the entire WIT in the signature then the signature will fail if the WIT is substituted and the attacker will not be able to replace the WIT.
+
+The last option to sign the WIT as part of the POP requires the least special verification steps. This is the approach taken by http message signatures in {{http-sig-auth}}. THe DPOP mechanism should take a similar approach if it continues to be part of the proposal.
```
jsalowey, line 545:
```suggestion
The last option to sign the WIT as part of the POP requires the fewest special verification steps. This is the approach taken by http message signatures in {{http-sig-auth}}. The DPOP mechanism should take a similar approach if it continues to be part of the proposal.
```

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -532,6 +532,18 @@ The proof of possession MUST be time limited. A PoP should only be valid over th
...
 
+* Binding WIT to Proof of Possession
+
+If the WIT itself is not bound in the proof of possession then it is possible that the WIT could be replaced by an attacker. If multiple WITs with different claims is issued for the same key this could result in the proof of possession having different meaning than what the sender intended and could give the attacker an advantage. Here are some possible mitigations.
+
+If the WIT only contains information that is already included in the POP signature then the different in WITs would not be meaningful. The verifier must check that the signed information matches the information in the WIT. Care must also be taken to understand what is substantial information. For example, even a unique identifier such as a jti claim could be problematic if the system associates additional information to this identifier.
+
+If the system ensures that the information associated with a WIT is the same for a given key that is embedded in the WIT then the attacker will not be able to substitute a different WIT with the signature. Here to care must be taken to understand what is substantial difference is.
```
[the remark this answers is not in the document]
jsalowey: I think it would just be simpler to include a hash of the WIT in the signature.  If not there will always be a chance that there is a substantive different between WITs that results in a security flaw.

yaronf:
Closing: added the WIT into WPT.

PR #58, opened 2024-08-26
Change trust root to trust anchor

kenmccracken-google:
Change trust root to trust anchor, to align with Trust Anchor terminology as used in rfc5280.


bc-pi (approved):
sure

Issue #39, opened 2024-07-04
Array of additional tokens


PieterKas:
Commenting as identity enthusiast as opposed to WIMSE co-chair.

The oth claim allows for a certain degree of extensibility, but raises a question about what to do if there is more  than one additional token for which a hash should be included. should this be an array of hashes? How would an implementor know which of the tokens is corresponding to which of the hashes?

bc-pi:
The `oth` claim as it's defined in -00 is kinda garbage, if we're being honest. 

Something more along the lines of what @yaronf suggested in https://github.com/yaronf/wimse-s2s/pull/19#discussion_r1648132681 that resulted in https://github.com/yaronf/wimse-s2s/issues/25 would probably be a better approach (to the extent we need this kind of extensibility anyway). While still aiming to avoid reinventing HTTP signatures...

yaronf:
Decided to avoid, mainly due to complexity. See also #25.

PR #60, opened 2024-09-13
More flexible `signed_headers` approach

### LINKED ISSUE #25: a construct that's both simple and extensible(?)
> Although we can definitely punt on it for now, I think we could choose a construct that's both simple and extensible.
>
>```json
>"signed_headers": {
>    "txn-token": "txn token hash",
>    "x-custom-context": "x-custom-context hash"
>}
>```
> Unfortunately we need the WPT to sign miscellaneous stuff, because we don't want a new WIT for every call. Having said that, as far as I can tell it does NOT need to sign the WIT itself.

_Originally posted by @yaronf in https://github.com/yaronf/wimse-s2s/pull/19#discussion_r1648132681_

arndt-s:
Introduces `signed_headers` as suggested by Yaron. 

This approach removes `tth` and `oth` claims for a more flexible `signed_headers` claim which can transfer proof for any (http) header.

Resolves #25 #39

### DIFF
--- a/draft-ietf-wimse-s2s-protocol.md
+++ b/draft-ietf-wimse-s2s-protocol.md
@@ -319,8 +319,8 @@ Optionally, the `Workload Proof Token` (WPT) may include the following, addition
      authorization context of the request. The value, as per {{Section 4.1 of RFC9449}},
      is the base64url encoding of the SHA-256 hash of the ASCII encoding of the access token's value.
     * `signed_headers`: A key-value map of headers where
-         * key: The name of the request header. In HTTP this is the HTTP-Header. In other protocols different headers can be used.
-         * value: base64url-encoded SHA-256 hash of the header value as present in the request.
+         * key: The name of the header field. In HTTP this can be generic, request or entity fields. This may differ in other protocols.
+         * value: base64url-encoded SHA-256 hash of the header value present in the request.
 
 An example WPT might look like the following:
 
@@ -409,9 +409,9 @@ To validate the WPT in the request, the recipient MUST ensure the following:
 
 ### `ath` claim vs `Authorization` key in `signed_header` claim
 
-Because access tokens are transmitted via the `Authorization` header it is possible to transport proof via `ath` claim, as an `Authorization` key in the `signed_headers` claim or both. The authors believe that supporting `ath` improves interoperability, as `ath` claim has been defined before by {{Section 4.1 of RFC9449}}.
+Because access tokens are transmitted via the `Authorization` header it is possible to transport proof via `ath` claim, as an `Authorization` key in the `signed_headers` claim or both. The authors believe that supporting `ath` improves interoperability, as it has been defined for similar reasons before by {{Section 4.1 of RFC9449}}. If both are present, both MUST be validated.
 
-It is important to keep in mind hat the processing between `ath` and `signed_headers` differs. `ath` would only compare the hashes of the actual token whilst the `Authorization` header in the `signed_headers` claim would compare the entire header value, including for example, a `Bearer ` prefix.
+It is important to keep in mind hat the processing rules between `ath` and `signed_headers` differs. `ath` would only compare the hashes of the actual token whilst the `Authorization` value in the `signed_headers` claim would compare the entire header value, including for example, a `Bearer ` prefix.
 
 ## Option 2: Authentication Based on HTTP Message Signatures {#http-sig-auth}
 
--- a/draft-ietf-wimse-s2s-protocol.md
+++ b/draft-ietf-wimse-s2s-protocol.md
@@ -313,17 +313,14 @@ A WPT contains the following:
     * `exp`: The expiration time of the WIT (as defined in {{Section 4.1.4 of RFC7519}}). WPT lifetimes MUST be short,
      e.g., on the order of minutes or seconds.
     * `jti`: A unique identifier for the token.
+
+Optionally, the `Workload Proof Token` (WPT) may include the following, additional claims, that restrict the context further:
     * `ath`: Hash of the OAuth access token, if present in the request, which might convey end-user identity and
      authorization context of the request. The value, as per {{Section 4.1 of RFC9449}},
      is the base64url encoding of the SHA-256 hash of the ASCII encoding of the access token's value.
-    * `tth`: Hash of the Txn-Token {{?I-D.ietf-oauth-transaction-tokens}}, if present in the request,
-     which might convey end-user identity and authorization context of the request. The value MUST be the result of
-     a base64url encoding (as defined in {{Section 2 of RFC7515}}) of the SHA-256 hash of
-     the ASCII encoding of the associated token's value.
-    * `oth`: Hash of any other token in the request that might convey end-user identity and authorization context of the
-     request. The value MUST be the result of a base64url encoding (as defined in {{Section 2 of RFC7515}}) of the
-     SHA-256 hash of the ASCII encoding of the associated token's value.
-     (note: this is less than ideal but seems we need something like this for extensibility)
+    * `signed_headers`: A key-value map of headers where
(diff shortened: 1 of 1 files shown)


draft-ietf-wimse-s2s-protocol.md
```diff
@@ -313,17 +313,14 @@ A WPT contains the following:
...
-    * `tth`: Hash of the Txn-Token {{?I-D.ietf-oauth-transaction-tokens}}, if present in the request,
-     which might convey end-user identity and authorization context of the request. The value MUST be the result of
-     a base64url encoding (as defined in {{Section 2 of RFC7515}}) of the SHA-256 hash of
-     the ASCII encoding of the associated token's value.
-    * `oth`: Hash of any other token in the request that might convey end-user identity and authorization context of the
-     request. The value MUST be the result of a base64url encoding (as defined in {{Section 2 of RFC7515}}) of the
-     SHA-256 hash of the ASCII encoding of the associated token's value.
-     (note: this is less than ideal but seems we need something like this for extensibility)
+    * `signed_headers`: A key-value map of headers where
+         * key: The name of the header field. In HTTP this can be generic, request or entity fields. This may differ in other protocols.
```
arndt-s, line 322: My thinking here is, that not restricting it to HTTP allows `WPT` to be used for example with `Kafka` or `gRPC headers`.

yaronf: Yes and no. The normative text should be focused on HTTP. If we add a note about a future non-HTTP variant of this protocol, it should be very clear that it's outside the normative "flow".

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -401,11 +401,17 @@ To validate the WPT in the request, the recipient MUST ensure the following:
 * Optionally, check that the value of the `jti` claim has not been used before in the time window in which the
  respective WPT would be considered valid.
 * If presented in conjunction with an OAuth access token, the value of the `ath` claim matches the hash of that token's value.
-* If presented in conjunction with a Txn-Token, the value of the `tth` claim matches the hash of that token's value.
-* If presented in conjunction with a token conveying end-user identity or authorization context, the value of
- the `oth` claim matches the hash of that token's value.
+* If `Workload-Proof-Token` contains `signed_headers` claim:
+   * The request contains a header for each claim in the `signed_headers`.
+   * Each corresponding value in the claim matches the hash of the value in the header.
+   * Headers not present in the `Workload-Proof-Token` are accepted without further proof.
```
yaronf, line 407: Huh? An attacker can add a header with the same name as an existing header and (depending on implementation) get it accepted.

I suggest to remove this bullet and change the first one to:
> The request contains exactly one header for each claim in `signed_headers` and no additional headers.

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -401,11 +401,17 @@ To validate the WPT in the request, the recipient MUST ensure the following:
...
+   * If a header is present more than once in the request, each value has a corresponding hash in the `Workload-Proof-Token`. Proof of just a single value in a multi-value header are not allowed.
```
yaronf, line 408: I believe (but don't consider myself an HTTP expert) that if a header is repeated multiple times, this is equivalent to concatenating the values together. Oh, here's the text from RFC 9421 again (@bc-pi will love it, I'm sure):

> Unless overridden by additional parameters and rules, HTTP field values MUST be combined into a single value as defined in [Section 5.2](https://rfc-editor.org/rfc/rfc9110#section-5.2) of [[HTTP](https://www.rfc-editor.org/rfc/rfc9421.html#RFC9110)] to create the component value. Specifically, HTTP fields sent as multiple fields MUST be combined by concatenating the values using a single comma and a single space as a separator ("," + " "). Note that intermediaries are allowed to combine values of HTTP fields with any amount of whitespace between the commas, and if this behavior is not accounted for by the verifier, the signature can fail, since the signer and verifier will see a different component value in their respective signature bases. For robustness, it is RECOMMENDED that signed messages include only a single instance of any field covered under the signature, particularly with the value for any list-based fields serialized using the algorithm below. This approach increases the chances of the field value remaining untouched through intermediaries. Where that approach is not possible and multiple instances of a field need to be sent separately, it is RECOMMENDED that signers and verifiers process any list-based fields taking all individual field values and combining them based on the strict algorithm below, to counter possible intermediary behavior.

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -401,11 +401,17 @@ To validate the WPT in the request, the recipient MUST ensure the following:
...
+
+### `ath` claim vs `Authorization` key in `signed_header` claim
 
+Because access tokens are transmitted via the `Authorization` header it is possible to transport proof via `ath` claim, as an `Authorization` key in the `signed_headers` claim or both. The authors believe that supporting `ath` improves interoperability, as it has been defined for similar reasons before by {{Section 4.1 of RFC9449}}. If both are present, both MUST be validated.
 
+It is important to keep in mind hat the processing rules between `ath` and `signed_headers` differs. `ath` would only compare the hashes of the actual token whilst the `Authorization` value in the `signed_headers` claim would compare the entire header value, including for example, a `Bearer ` prefix.
```
yaronf, line 414:
```suggestion
It is important to keep in mind that the processing rules between `ath` and `signed_headers` differ. `ath` would only compare the hashes of the actual token whilst the `Authorization` value in the `signed_headers` claim would compare the entire header value, including for example, a `Bearer ` prefix.
```

yaronf (changes requested):

draft-ietf-wimse-s2s-protocol.md
bc-pi: I'm not sure I know what exactly "generic, request or entity fields" means exactly in HTTP but I'm pretty sure it's not what should be here (especially entity).

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -401,11 +401,17 @@ To validate the WPT in the request, the recipient MUST ensure the following:
 * Optionally, check that the value of the `jti` claim has not been used before in the time window in which the
  respective WPT would be considered valid.
 * If presented in conjunction with an OAuth access token, the value of the `ath` claim matches the hash of that token's value.
-* If presented in conjunction with a Txn-Token, the value of the `tth` claim matches the hash of that token's value.
-* If presented in conjunction with a token conveying end-user identity or authorization context, the value of
- the `oth` claim matches the hash of that token's value.
+* If `Workload-Proof-Token` contains `signed_headers` claim:
```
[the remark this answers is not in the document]
bc-pi:
```suggestion
* If Workload-Proof-Token contains a `signed_headers` claim:
```

draft-ietf-wimse-s2s-protocol.md
bc-pi: I must admit to being pretty confused at this stage.

draft-ietf-wimse-s2s-protocol.md
bc-pi: This solution, which shouldn't attempt to be a generic signature thing, should probably constrain itself to single value things.

bc-pi (changes requested):
We probably should discuss this a bit more on our call, again, before digging into this more.

draft-ietf-wimse-s2s-protocol.md
jsalowey: I'm in agreement that we want to focus the POP mechanism on HTTP. Trying to anticipate some other protocol or constraints at this point is difficult. We will need to define the POP for other protocols in other documents.  I think there is value in having the POP be flexible.  I think that is an argument in favor of HTTPSigs.  
Just to mention here It would be great if we keep WIT as not tightly bound to HTTP as it is now.

Issue #46, opened 2024-07-04
X.509 fields


PieterKas:
Commenting as identity enthusiast as opposed to WIMSE co-chair

Section 5 states that the identifier may be used in an X.509 certificate. From reading the draft, this may be a client or a server certificate. Should we included additional details about the certificates themeselves? Should a workload be able to use the same certificate for client authentication and as a server? What are the extended key usage parameters that should be supported (can a cert have both)?

jsalowey:
PR#63 adds some of this.  We may want to make more specification around what can be in the URI (no port, username; trust domain lowercase)

jsalowey:
This issue is addressed by text in section 5 an section 3.

PR #63, opened 2024-10-09
rework of the tls section

jsalowey:
Reworked the TLS section a little.  mostly addresses issue #46


draft-ietf-wimse-s2s-protocol.md
```diff
@@ -486,25 +486,21 @@ A signed response would be:
...
 
 Servers wishing to use the WIMSE certificate for authorizing the client MUST require client certificate authentication in the TLS handshake. Other methods of post handshake authentication are not specified by this document.
 
-WIMSE clients and servers MUST validate that the trust domain portion of the WIMSE certificate matches the expected trust domain for the other side of the connection.
+WIMSE server certificates MUST have the id-kp-serverAuth extended key usage {{!RFC5280}} field set and WIMSE client certificates MUST have the id-kp-clientAuth extended key usage field set. A certificate that is used for both client and server connections may has both fields set. This specification does not make any other requirements beyond {{!RFC5280}} on the contents of WIMSE certificates or on the certificates that issue WIMSE certificates.
```
yaronf, line 495: Asking out of ignorance: in the past, EKU was hit-and-miss. Is it now implemented so widely that we can have it as a MUST?

jsalowey: I think its probably still a little hit and miss.  I think we can try to raise the bar here, but maybe its not really all that important enough to make it a barrier. I'd be happy with a SHOULD.

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -486,25 +486,21 @@ A signed response would be:
 
 # Using Mutual TLS for Service To Service Authentication {#mutual-tls}
 
-The WIMSE workload identity may be carried within an X.509 certificate. The WIMSE workload identity MUST be encoded in a SubjectAltName extension of type URI.  There MUST be only one SubjectAltName extension of type URI in a WIMSE certificate.  The WIMSE certificate may contain SubjectAltName extensions of other types such as DNSName.
+The WIMSE workload identity may be carried within an X.509 certificate. The WIMSE workload identity MUST be encoded in a SubjectAltName extension of type URI.  There MUST be only one SubjectAltName extension of type URI in a WIMSE certificate.  If the workload will act as a TLS server for clients that do not understand WIMSE workload identities it is RECOMMENDED that WIMSE certificate contain a SubjectAltName of DNSName with the appropriate DNS names for the server. The certificate may contain SubjectAltName extensions of other types.
```
yaronf, line 489:
```suggestion
The WIMSE workload identity may be carried within an X.509 certificate. The WIMSE workload identity MUST be encoded in a SubjectAltName extension of type URI.  There MUST be only one SubjectAltName extension of type URI in a WIMSE certificate.  If the workload will act as a TLS server for clients that do not understand WIMSE workload identities it is RECOMMENDED that WIMSE certificate contain a SubjectAltName of type DNSName with the appropriate DNS names for the server. The certificate may contain SubjectAltName extensions of other types.
```

draft-ietf-wimse-s2s-protocol.md
yaronf, line 495:
```suggestion
WIMSE server certificates SHOULD have the id-kp-serverAuth extended key usage {{!RFC5280}} field set and WIMSE client certificates SHOULD have the id-kp-clientAuth extended key usage field set. A certificate that is used for both client and server connections may have both fields set. This specification does not make any other requirements beyond {{!RFC5280}} on the contents of WIMSE certificates or on the certification authorities that issue WIMSE certificates.
```

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -486,25 +486,21 @@ A signed response would be:
...
 
-## Host Name Validation
+## Server Name Validation {#server-name}
 
-<cref>TODO: need to define trust anchor used to validate the certificate is appropriate for the trust domain.</cref>
+If the WIMSE client uses a hostname to connect to the server and the server certificate contain a DNS SAN the client MUST perform standard host name validation ({{Section 6.3 of RFC9525}}) unless it is configured with the information necessary to validate the peer's WIMSE identity. If the client did not perform standard host name validation then WIMSE client SHOULD further use the WIMSE workload identifier to validate the server.  The host portion of the WIMSE URI is NOT treated as a host name as specified in section 6.4 of {{!RFC9525}} but rather as a trust domain. The server identity is encoded in the path portion of the WIMSE workload identifier in a deployment specific way. Validating the WIMSE workload identity could be a simple match on the trust domain and path portions of the identifier or validation may be based on the specific details on how the identifier is constructed. The path portion of the WIMSE identifier MUST always be considered in the scope of the trust domain.
```
yaronf, line 499:
```suggestion
If the WIMSE client uses a hostname to connect to the server and the server certificate contain a DNS SAN the client MUST perform standard host name validation ({{Section 6.3 of RFC9525}}) unless it is configured with the information necessary to validate the peer's WIMSE identity. If the client did not perform standard host name validation then the WIMSE client SHOULD further use the WIMSE workload identifier to validate the server.  The host portion of the WIMSE URI is NOT treated as a host name as specified in section 6.4 of {{!RFC9525}} but rather as a trust domain. The server identity is encoded in the path portion of the WIMSE workload identifier in a deployment specific way. Validating the WIMSE workload identity could be a simple match on the trust domain and path portions of the identifier or validation may be based on the specific details on how the identifier is constructed. The path portion of the WIMSE identifier MUST always be considered in the scope of the trust domain.
```

yaronf (changes requested):
Mostly editorial changes, but also MUST->SHOULD

draft-ietf-wimse-s2s-protocol.md
jsalowey: Resolved by Yaron's suggestion.

PR #66, opened 2024-10-11
Add 'wth' claim to WPT

arndt-s:
Adds 'wth' claim to WPT to bind WPT to WITs.


draft-ietf-wimse-s2s-protocol.md
```diff
@@ -313,6 +313,8 @@ A WPT contains the following:
     * `exp`: The expiration time of the WIT (as defined in {{Section 4.1.4 of RFC7519}}). WPT lifetimes MUST be short,
      e.g., on the order of minutes or seconds.
     * `jti`: A unique identifier for the token.
+    * `wth`: Hash of the Workload Identity Token. The value, as defined in {{to-wit}}, is the base64url encoding of the SHA-256
+     hash of the ASCII encoding of the token's value.
```
yaronf, line 317:
```suggestion
    * `wth`: Hash of the Workload Identity Token, defined in {{to-wit}}. The value is the base64url encoding of the SHA-256
     hash of the ASCII encoding of the token's value.
```

The token is defined in {{to-wit}}, not the hash.

yaronf (changes requested):
Editorial

draft-ietf-wimse-s2s-protocol.md
jsalowey: Yaron's suggestion looks good to me.

yaronf:
@arndt-s please also update the relevant IANA section so we don't forget.

yaronf (approved):

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -351,11 +354,12 @@ The decoded JWT claims of the WPT from the example above are shown here:
...
- "iss": "wimse://example.com/specific-workload",
- "aud": "https://service.example.com/path",
- "exp": 1717612820,
- "jti": "__bwc4ESC3acc2LTC1-_x",
- "ath": "CL4wjfpRmNf-bdYIbYLnV9d5rMARGwKYE10wUwzC0jI"
+  "ath": "CL4wjfpRmNf-bdYIbYLnV9d5rMARGwKYE10wUwzC0jI",
+  "aud": "https://service.example.com/path",
+  "exp": 1728658672,
+  "iss": "wimse://example.com/specific-workload",
+  "jti": "4b42c5f611e2b1cfa1d2c41b3a2fb782",
```
bc-pi, line 361: Whhhhhhyyyyyyyy?  Hex in jti values is so aesthetically displeasing.

bc-pi:
I realize I missed a meeting but why did we add this?

bc-pi:
What's the threat? Different WITs with the same cnf key and a compromised middle thing?

bc-pi:
I just don't see it...

bc-pi:
Iss in the WPT is even less useful with this too.

draft-ietf-wimse-s2s-protocol.md
arndt-s: If you feel strongly about this I can provide a new PR that changes this identifier ;)

arndt-s:
> What's the threat? Different WITs with the same cnf key and a compromised middle thing?

Yes, currently we don't have strong binding between WPT and WIT. Only the fact that `WIT sub == WIT iss`.

> Iss in the WPT is even less useful with this too.

I don't understand the "even less" part but I see that this is somewhat duplicated now.

yaronf:
> Yes, currently we don't have strong binding between WPT and WIT. Only the fact that WIT sub == WIT iss.

Should be: WIT sub == WPT iss.

And just to state it explicitly: people are likely to extend the WIT with various "junk" (i.e., policy) and that wouldn't be covered otherwise.

draft-ietf-wimse-s2s-protocol.md
bc-pi: I do kinda. `__bwc4ESC3acc2LTC1-_x` was okay. and the order of the claims was more readable before

PR #67, opened 2024-10-13
Change log for draft -01

yaronf:


bc-pi:
We should do these entries in conjunction with the changes...

arndt-s (approved):

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -586,6 +586,16 @@ TODO: `Workload-Proof-Token` from {{dpop-esque-auth}}
 # Document History
 <cref>RFC Editor: please remove before publication.</cref>
 
+## draft-ietf-wimse-s2s-protocol-01
+
+* Addressed multiple comments from Pieter.
```
bc-pi, line 591:
```suggestion
* Addressed multiple comments from Pieter, co-chair of the WIMSE WG but comments made as an individual.  
```

yaronf: I already merged the PR, will add this text separately.

bc-pi: race condition

bc-pi: it's ok without too

PR #76, opened 2024-11-07
Add Option 3 WIT with Authentication Based on the Transport-layer

### LINKED ISSUE #75: WIT with cnf referencing Transport-layer security
This spec should support an Option that builds on secure Workload Transport-layer authentication.  Well-established protocols exist for Workload-to-Workload communication over mutually authenticated TLS 1.3.

I would like to propose an `Option 3: Transport-layer Proof-of-possession`, with a reference to the TLS credential used in the `cnf` claim of the WIT, in the manner described in https://datatracker.ietf.org/doc/html/rfc8705#section-3.1

### The Workload Identity Token {#to-wit}

_Change the description of `cnf` to the following:_

* `cnf` : A confirmation claim containing either the public key of the workload OR a cryptographically strong hash of the X.509 certificate conveyed during Transport layer security. 
    * The public key `cnf` contains the jwk member as defined in {{Section 3.2 of RFC7800}}. The workload MUST prove possession of the corresponding private key when presenting the WIT to another party, which can be accomplished by using it in conjunction with one of the methods in {{dpop-esque-auth}} or {{http-sig-auth}}. As such, it MUST NOT be used as a bearer token and is not intended for use in the Authorization header.
    * The X.509 certificate hash `cnf` contains the `x5t#S256` hash as described in [Section 3.1](https://datatracker.ietf.org/doc/html/rfc8705#section-3.1) of [[RFC8705](https://datatracker.ietf.org/doc/html/rfc8705)]. When this value is provided, Proof-of-possession is delegated to the Transport-layer. The receiving Workload MUST verify that the `x5t#S256` claim matches the hash of the Client's X.509 Certificate presented at the Transport-layer.

### Option 3: Transport-layer Proof-of-possession

In the case where Workload-to-Workload communication is secured by Mutually-authenticated [Transport Layer Security 1.3](https://datatracker.ietf.org/doc/html/rfc8446#section-2), Proof-of-possession is conveyed via each Workload's `CertificateVerify` message. In this case, the Transport-layer guarantees that each Workload has already validated its Peer's Proof-of-possession of the Private Key. Proof-of-possession is conveyed during the TLS 1.3 handshake protocol using each peer's `Certificate` and `CertificateVerify` messages, along with other protocol messages transmitted. The Transport-layer must cache the SHA-256 hash of the validated Client `Certificate` and the first Subject Alternative Name (SAN) URI entry for use at the Application-layer. When the Destination Workload receives a WIT at the Application-layer with `cnf` claim with `x5t#S256` property, it MUST establish Proof-of-possession as follows:
* Validate that the Certificate-bound WIT's `cnf` claim's `x5t#S256` property matches the cached SHA-256 hash of the validated Client Certificate.
* Validate that the WIT's `sub` claim matches the first Subject Alternative Name (SAN) URI entry.

The first step proves that the WIT was issued to the holder of the PrivateKey for the X.509 Certificate presented by the client, and that Private Key was used to establish the mTLS session over which the WIT was transmitted. The second step validates that both the X.509 Certificate and the WIT were issued to the same Workload.

kenmccracken-google:
In this approach, the WIT refers to the mTLS X.509 Certificate credential for establishing Workload Authentication.
Resolves #75

### DIFF
--- a/draft-ietf-wimse-s2s-protocol.md
+++ b/draft-ietf-wimse-s2s-protocol.md
@@ -176,7 +176,8 @@ While the URI encoding rules allow host names to be specified as IP addresses, I
 
 As noted in the Introduction, for many deployments communication between workloads cannot use
 end-to-end TLS. For these deployment styles, this document proposes application-level protections.
-For deployments using end-to-end TLS, application-level credentials may enrich the security context.
+For deployments using end-to-end TLS, application-level credentials may be used to enrich the 
+application security context.
 
 The current version of the document includes three alternatives, all using the newly introduced
 Workload Identity Token ({{to-wit}}). The first alternative ({{dpop-esque-auth}}) is inspired by the OAuth DPoP specification.
@@ -189,7 +190,7 @@ A comparison of the first two alternatives is attempted in {{app-level-compariso
 ## The Workload Identity Token {#to-wit}
 
 The Workload Identity Token (WIT) is a JWS {{RFC7515}} signed JWT {{RFC7519}} that represents the identity of a workload.
-It is issued by the Identity Server and binds a public key to the workload identity.
+It is issued by the Identity Server and binds a public key or certificate to the workload identity.
 A WIT MUST contain the following:
 
 * in the JOSE header:
@@ -202,9 +203,9 @@ A WIT MUST contain the following:
     * `exp`: The expiration time of the token (as defined in {{Section 4.1.4 of RFC7519}}).
       WITs should be refreshed regularly, e.g. on the order of hours.
     * `jti`: A unique identifier for the token.
-    * `cnf`: A confirmation claim containing either the public key of the workload OR a cryptographically strong hash of the X.509 certificate that was conveyed during Transport layer security.
-        * The public key `cnf` contains the `jwk` member as defined in {{Section 3.2 of RFC7800}}. The workload MUST prove possession of the corresponding private key when presenting the WIT to another party, which can be accomplished by using it in conjunction with one of the methods in {{dpop-esque-auth}} or {{http-sig-auth}}. As such, it MUST NOT be used as a bearer token and is not intended for use in the `Authorization` header.
-        * The X.509 certificate hash `cnf` contains the `x5t#S256` hash as described in [Section 3.1](https://datatracker.ietf.org/doc/html/rfc8705#section-3.1) of [[RFC7805](https://datatracker.ietf.org/doc/html/rfc8705)]. When this value is provided, Proof-of-possession is delegated to the Transport-layer using the method in {{transport-layer-pop}}. The receiving Workload MUST verify that the `x5t#S256` claim matches the hash of the Client's X.509 Certificate presented at the Transport-layer.
+    * `cnf`: A confirmation claim containing either the public key of the workload OR a the hash of the X.509 certificate of the workload.
+        * The `cnf` MAY contain the `jwk` member as defined in {{Section 3.2 of RFC7800}}. The workload MUST prove possession of the corresponding private key when presenting the WIT to another party, which can be accomplished by using it in conjunction with one of the methods in {{dpop-esque-auth}} or {{http-sig-auth}}. As such, it MUST NOT be used as a bearer token and is not intended for use in the `Authorization` header.
+        * The `cnf` MAY contain the `x5t#S256` member as defined in [Section 3.1](https://datatracker.ietf.org/doc/html/rfc8705#section-3.1) of [[RFC7805](https://datatracker.ietf.org/doc/html/rfc8705)]. When this value is provided, Proof-of-possession is delegated to the Transport layer using the method described in {{transport-layer-pop}}. The receiving Workload MUST verify that the `x5t#S256` claim matches the SHA-256 hash of the Client's X.509 Certificate presented at the Transport layer.
 
 An example WIT might look like this (all examples, of course, are non-normative and with line breaks and extra space for readability):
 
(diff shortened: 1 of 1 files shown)


bc-pi:
My perception here is that this additional option and/or the rational behind it isn't well understood by the WG or the editors/authors of this document. Nor does there seem to be much interest in perusing it. I understand that doing a PR was suggested to help facilitate discussion. But I would humbly suggest that a PR that adds oneself as an author and breaks the document build process is not a great way to solicit engagement.

kenmccracken-google:
Hi Brian, Thanks for your feedback.  I apologize for the build failures.  I'll try to get them fixed.

I tried the build from a couple of different machines but it doesn't seem to work for me, including just building a fresh clone of https://github.com/ietf-wg-wimse/draft-ietf-wimse-s2s-protocol.git .  I'll see if I can figure out the problem, as of now I consistently get a failure from a package called `kramdown-rfc`, which I am not familiar with.  I've attached one of the outputs from a Mac, but I got a similar failure with the same module from a Debian linux machine.  I tried following https://github.com/martinthomson/i-d-template/blob/main/doc/SETUP.md.

[failures.txt](https://github.com/user-attachments/files/17894252/failures.txt)

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -43,6 +43,10 @@ author:
     fullname: "Yaron Sheffer"
     organization: Intuit
     email: "yaronf.ietf@gmail.com"
+ -
+    fullname: "Ken McCracken"
+    organization: Google
+    email: "kenmccracken@google.com"
```
jricher, line 49: IETF Process note - documents should only have at most 5 authors. Other contributors can be listed in the acknowledgements section. It should be fine to add more for now but this will need to be trimmed back prior to ultimate publication.

jricher:
@bc-pi @kenmccracken-google The build process is broken due to the linter picking up trailing spaces on two lines (see below). I ran `make fix-lint` and was able to build the PR locally, if this is to be included this would need to be fixed before merge:

``` diff
[jricher@Boudica ~/Projects/draft-ietf-wimse-s2s-protocol (mtls-workload-proof*) ]$ git diff
diff --git a/draft-ietf-wimse-s2s-protocol.md b/draft-ietf-wimse-s2s-protocol.md
index a2df4e1..0824754 100644
--- a/draft-ietf-wimse-s2s-protocol.md
+++ b/draft-ietf-wimse-s2s-protocol.md
@@ -176,13 +176,13 @@ While the URI encoding rules allow host names to be specified as IP addresses, I
 
 As noted in the Introduction, for many deployments communication between workloads cannot use
 end-to-end TLS. For these deployment styles, this document proposes application-level protections.
-For deployments using end-to-end TLS, application-level credentials may be used to enrich the 
+For deployments using end-to-end TLS, application-level credentials may be used to enrich the
 application security context.
 
 The current version of the document includes three alternatives, all using the newly introduced
 Workload Identity Token ({{to-wit}}). The first alternative ({{dpop-esque-auth}}) is inspired by the OAuth DPoP specification.
 The second alternative ({{http-sig-auth}}) is based on the HTTP Message Signatures RFC. The third
-alternative ({{transport-layer-pop}}) is based on the TLS 1.3 and Token Binding RFCs. 
+alternative ({{transport-layer-pop}}) is based on the TLS 1.3 and Token Binding RFCs.
 We present the alternatives and expect
 the working group to select those that should progress towards IETF consensus.
 A comparison of the first two alternatives is attempted in {{app-level-comparison}}.
```

draft-ietf-wimse-s2s-protocol.md
kenmccracken-google: Moved my name to Contributors list at the bottom.

yaronf:
And now back to the matter at hand. Following Ken's post to the mailing list I think I understand the motivation somewhat better, but I still don't think this is a good idea.

The WIT is not an access token. It's a signed token that incorporates workload-specific information (mainly the workload's ID and its public key, but potentially future extensions) and therefore it should normally only be shared between a workload and its immediate neighbors (its successors on the call chain). As such, there are two options:
- The next hop is protected with MTLS (implicitly, there is no middlebox to break TLS security): no need for WIT because the identity information and public key are provided in the certificate.
- The next hop is not protected with MTLS but rather with server-authenticated TLS (and potentially a middlebox): we send the WIT over the network hop, and bind it to a signature or proof-of-possession.

So I don't see a reason to combine the WIT with MTLS.

By the way, the WIT is "authenticated" - it is signed by a trusted party, the server that issued it.

kenmccracken-google:
I think it's important to allow for a baseline architecture for Service to Service Authentication that utilizes mTLS streams established using X.509 Certificate credentials.

| The WIT is not an access token. It's a signed token that incorporates workload-specific information (mainly the workload's ID and its public key, but potentially future extensions) and therefore it should normally only be shared between a workload and its immediate neighbors (its successors on the call chain). 

I understood the WIT to be a Workload Identity Token.  I agree it should contain the identity of the workload, represented by a URI, in the "sub" claim.  When it comes to proof-of-possession semantics under [RFC 7800, Section 3.1](https://datatracker.ietf.org/doc/html/rfc7800#section-3.1), I think the "cnf" claim should be used to identify the proof-of-possession key.  

In this context, I think that referencing an X.509 Certificate (that contains the Workload's public key) using "x5t#S256" should be an option for "cnf", as an alternative to inlining the public key in the "jwk" member.

| As such, there are two options:
| 
| The next hop is protected with MTLS (implicitly, there is no middlebox to break TLS security): no need for WIT because the identity information and public key are provided in the certificate.
| The next hop is not protected with MTLS but rather with server-authenticated TLS (and potentially a middlebox): we send the WIT over the network hop, and bind it to a signature or proof-of-possession.

I think from this description you would classify a proxy as a middlebox.  I think the middlebox proxy may be capable of hop-to-hop mTLS.  In this case, we would have the additional option:

* next hop is protected with mTLS, and is a middlebox.  The middlebox is a proxy that forwards requests to destination workloads over mTLS.

In this architecture, the middlebox must perform proof-of-possession verification of the Workload's private key.  The `Workload-Identity-Token` header can be forwarded intact to the destination Workload.  The destination Workload verifies the identity of the middlebox over mTLS.   

| By the way, the WIT is "authenticated" - it is signed by a trusted party, the server that issued it.

Agree, I think I wasn't clear in my wording.  You're correct, the WIT is signed by the issuer.  I probably should have said something like:

I think the Identity Server should issue WITs with a confirmation claim indicating the Workload's public key.  I think recipients of the WIT should require proof-of-possession of the Workload's private key, according to the information in the "cnf" claim.  I think the document should support verifying the proof-of-possession of the Workload's private key via one of 
1. WPT, as described in [1] Section 4.2.
2. HTTP Message signatures, as described in [1] Section 4.3.
3. mTLS, as described in [3].

When the Identity Server puts an "x5t#S256" value in the "cnf" claim, the proof-of-posession verification method follows the semantics of [RFC 8705, Section 3.1](https://datatracker.ietf.org/doc/html/rfc8705#name-jwt-certificate-thumbprint-).

yaronf:
Closing as outdated. See #75.

PR #81, opened 2024-11-29
Use rfcfold for section 4.2 and add make target

arndt-s:
See title


yaronf (approved):
Works for me. Thanks!

Issue #53, opened 2024-07-17
Use of WIT without proof of possession and interop with legacy systems


jsalowey:
The current draft says that the WIT must not be accepted without proof of possession of the corresponding private key.  This is for security reasons because if you are willing to accept the WIT without proof of possession then the security benefit of the WIT is greatly diminished.  

This restriction makes it difficult to integrate a WIT seamlessly into existing applications.  I question whether this could be done securely, but I think we should have the discussion:
1. Is there a sensible policy that would make this OK?  It seems that legacy application would be none the worse off, but application that could handle the WIT would still benefit.  I think there is a difficulty here in that you have client an services sides that might be updated a different times making it more complex.
2. Perhaps there is an alternate form of POP that would be easier.  For example binding to a TLS key or identity.  While this seems attractive, one of the reasons for the WIT is to suport cases where MTLS isn't realistic.

yaronf:
And then the do-nothing alternative:

3. Don't allow using WIT without PoP (granted there's no way to enforce this restriction in the real world), and rely on token exchange to translate between legacy bearer tokens and WITs.

kenmccracken-google:
I agree with both points (1.) and (2.).  
1.  is needed for backward-compatibility with existing "legacy" systems.  
2. allows WITs to be built on top of existing X.509-credentials, which should be part of the recommended mTLS solution.  

For both of these cases, claims within the WIT can further constrain access using "aud"/"scp" to express least-privileges access on top of an identity's (possibly-broad) access privileges imparted by Workload identity alone.  The "aud"/"scp" constraints can also be preserved during federation with OIDC-like IdPs.

IMO, the WIMSE s2s protocol should provide the option 2 for PoP, where the protocol can accept WITs layered on top of existing X.509-credentials for situations where mTLS *is* realistic.   In these cases, reducing the number of tokens required can reduce developer integration bugs and misconfigurations, when compared to the multi-token-requirements of Section 4.  X.509 Certificate hash binding or SAN identity binding should be considered, as an adaptation of https://www.ietf.org/archive/id/draft-ietf-wimse-arch-01.html#name-identity-credentials "Alternatively, sender constrained tokens can be used such as TLS session binding."

yaronf:
* For legacy systems, the bearer token's format is unspecified, so the WIMSE token cannot be coerced to have it.
* Also the type field would mismatch, and aud would be missing.

[...]

PR #84, opened 2024-12-23
coexistence with bearer tokens

jsalowey:
Resolve with Issue #53

### DIFF
--- a/draft-ietf-wimse-s2s-protocol.md
+++ b/draft-ietf-wimse-s2s-protocol.md
@@ -529,7 +529,8 @@ The WIT and WPT define new HTTP headers. They can therefore be presented along w
 property allows for transition from mechanisms using identity tokens based on bearer JWTs to proof of possession based WITs.
 A workload may implement a policy that accepts both bearer tokens and WITs during a transition period. This policy may be configurable
 per-caller to allow the workload to reject bearer tokens from callers that support WITs. Once a deployment fully supports WITs, then the use of
-bearer tokens for identity can be disabled through policy.
+bearer tokens for identity can be disabled through policy.  Implementations should be careful when implementing such a transition strategy,
+since the decision which token to prefer is made when the caller's identity has still not been authenticated, and needs to be revalidated following the authentication step.
 
 The WIT can also coexist with tokens used to establish security context, such as transaction tokens {{?I-D.ietf-oauth-transaction-tokens}}. In this case a workload's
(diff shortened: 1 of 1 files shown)


draft-ietf-wimse-s2s-protocol.md
```diff
@@ -523,6 +523,19 @@ so it cannot be easily reused in another context, which is often a security impr
 be designed to include extensibility to sign other fields, but this would make it closer to
 trying to reinvent Message Signatures.
 
+## Coexistence with JWT Bearer Tokens {coexist}
+
+The WIT and WPT define new HTTP headers. They can therefore be presented along with existing headers used for JWT bearer tokens. This
+property allows for transition from mechanisms using identity tokens based on bearer JWTs to proof of possession based WITs.
+A workload may implement a policy that accepts both bearer tokens and WITs during a transition period. This policy may be configurable
+per-caller to allow the workload reject bearer tokens from callers that support WITs. Once a deployment fully supports WITs, then the use of
```
yaronf, line 531:
```suggestion
per-caller to allow the workload to reject bearer tokens from callers that support WITs. Once a deployment fully supports WITs, then the use of
```

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -523,6 +523,19 @@ so it cannot be easily reused in another context, which is often a security impr
...
+bearer tokens for identity can be disabled through policy.
+
+The WIT can also coexist with tokens used to establish security context, such as transaction tokens {{?I-D.ietf-oauth-transaction-tokens}}. In this case a workload's
+authorization policy may take into account both the sending workload's identity and the information in the context token. For example, the
+identity in the WIT may be used to establish which API calls can be made and information in the context token may be used to determine
+which specific resources can be accessed.
+
```
yaronf, line 538: This is OK for now, but I think eventually we will want one subsection for coexistence with legacy stuff and another for authorization.

draft-ietf-wimse-s2s-protocol.md
yaronf: Suggest to add: Implementations should be careful when implementing such a transition strategy, since the decision which token to prefer is made when the caller's identity has still not been authenticated, and needs to be revalidated following the authentication step.

yaronf (changes requested):

yaronf (approved):

PR #98, opened 2025-02-27
Remove iss claim from WPT

### LINKED ISSUE #92: "iss" claim in WPT matching
Opened as working group member, not chair

From section 4.2 "The iss claim of the WPT matches the sub claim of the WIT.  (Note: not sure iss in the WPT is useful or necessary.)"

If the purpose of comparing the WPT "iss" claim and the WIT "sub" claim is to link the signature to the WIT, it seems that verifying the "wth" claim would have the same effect as long as the hash calculation includes the WIT "sub" claim. 

Are there other reasons why the "iss" claim should be verified?

arndt-s:
See title. Resolves #92

### DIFF
--- a/draft-ietf-wimse-s2s-protocol.md
+++ b/draft-ietf-wimse-s2s-protocol.md
@@ -337,8 +337,6 @@ A WPT contains the following:
     * `typ`: the WPT is explicitly typed, as recommended in {{Section 3.11 of RFC8725}},
      using the `application/wimse-proof+jwt` media type.
 * in the JWT claims:
-    * `iss`: The issuer of the token, which is the calling workload, represented by the same value as the `sub` claim
-     of the associated WIT.
     * `aud`: The audience of the token contains the HTTP target URI ({{Section 7.1 of RFC9110}}) of the request
      to which the WPT is attached, without query or fragment parts.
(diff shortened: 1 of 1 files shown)


yaronf (approved):

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -389,7 +387,6 @@ The decoded JWT claims of the WPT from the example above are shown here:
   "ath": "CL4wjfpRmNf-bdYIbYLnV9d5rMARGwKYE10wUwzC0jI",
   "aud": "https://workload.example.com/path",
   "exp": 1728658672,
-  "iss": "wimse://example.com/specific-workload",
```
bc-pi, line 392: do you also want to remove it from the example jwt above (lines ~362-370)? Probably should but I know that starts to pull the thread of if the signature is valid etc.

arndt-s: I'm regenerating the examples in a separate PR. Makes it easier to merge, etc.

bc-pi: okay, that works too

bc-pi (approved):

Issue #15, opened 2024-06-13
Authorization header and error responses


yaronf:
Examples in the the Message Signatures section do not include an Authorization header. But for a 401 response (which is likely if the workload-to-workload call is disallowed by policy) the spec requires WWW-Authenticate headers to be included. How can we return a 401, or do we need a whole new HTTP status code?

yaronf:
Keeping this issue open, see @arndt-s comment on #101 .

PR #101, opened 2025-03-01
add error section

jsalowey:
Resolves issue #15

### DIFF
--- a/draft-ietf-wimse-s2s-protocol.md
+++ b/draft-ietf-wimse-s2s-protocol.md
@@ -544,8 +544,6 @@ Errors may occur during the processing of the message signature or WPT. If the s
 such as an invalid signature, an expired validity time window, or a malformed data structure, an error is returned. Typically,
 this will be in response to an API call, so an HTTP status code such as 400 (Bad Request) is appropriate. This response could
 include more details as per {{RFC9457}}, such as an indicator that the wrong key material or algorithm was used.
-If the WPT or message signature is used for another purpose, such as authentication to start a session, a different response may
-be more appropriate, such as an HTTP status code of 401 (Unauthorized) or 403 (Forbidden).
 
 ## Coexistence with JWT Bearer Tokens {#coexist}
 
--- a/draft-ietf-wimse-s2s-protocol.md
+++ b/draft-ietf-wimse-s2s-protocol.md
@@ -536,6 +536,15 @@ so it cannot be easily reused in another context, which is often a security impr
 be designed to include extensibility to sign other fields, but this would make it closer to
 trying to reinvent Message Signatures.
 
+## Error Conditions
+
+Errors may occur during the processing of the message signature or WPT. If the signature verification fails for any reason,
+such as an invalid signature, an expired validity time window, or a malformed data structure, an error is returned. Typically,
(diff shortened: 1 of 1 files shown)


draft-ietf-wimse-s2s-protocol.md
```diff
@@ -536,6 +538,15 @@ so it cannot be easily reused in another context, which is often a security impr
 be designed to include extensibility to sign other fields, but this would make it closer to
 trying to reinvent Message Signatures.
 
+## Error Conditions
+
+Errors may occur during the processing of the message signature or WPT. If the signature verification fails for any reason,
+such as an invalid signature, an expired validity time window, or a malformed data structure, an error is returned. Typically,
+this will be in response to an API call, so an HTTP status code such as 400 (Bad Request) is appropriate. This response could
+include more details as per {{RFC7807}} and {{RFC9457}}, such as an indicator that the wrong key material or algorithm was used.
```
yaronf, line 546:
```suggestion
include more details as per {{RFC9457}}, such as an indicator that the wrong key material or algorithm was used.
```

RFC 9457 obsoletes RFC 7807.

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -51,6 +51,8 @@ normative:
 
 informative:
   IANA.JOSE.ALGS: IANA.jose_web-signature-encryption-algorithms
+  RFC7807:
```
yaronf, line 54:
```suggestion
```

See below.

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -536,6 +538,15 @@ so it cannot be easily reused in another context, which is often a security impr
...
 trying to reinvent Message Signatures.
 
+## Error Conditions
+
+Errors may occur during the processing of the message signature or WPT. If the signature verification fails for any reason,
+such as an invalid signature, an expired validity time window, or a malformed data structure, an error is returned. Typically,
+this will be in response to an API call, so an HTTP status code such as 400 (Bad Request) is appropriate. This response could
+include more details as per {{RFC7807}} and {{RFC9457}}, such as an indicator that the wrong key material or algorithm was used.
+If the WPT or message signature is used for another purpose, such as authentication to start a session, a different response may
+be more appropriate, such as an HTTP status code of 401 (Unauthorized) or 403 (Forbidden).
```
yaronf, line 548: This sentence is problematic IMO for two reasons:

- We don't define the concept of a "session", rather we assume a classic REST model where each request is independent of the others. And of course each WPT or message signature provides workload authentication.
- This still does not address my question in #15 regarding the `Authorization` header.

yaronf (changes requested):

draft-ietf-wimse-s2s-protocol.md
jsalowey: I'm happy to remove that part of the sentence.  I agree that it is problematic.  I don't think we can put anything in the authorization header as we do not use and leave it reserved for another token.

arndt-s:
I think we should have added some text saying why 401 is not suitable, otherwise it'll leave the reader just guessing.

PR #125, opened 2025-05-23
Clarification on conditional claims

### LINKED ISSUE #91: Editorial: State Required vs Optional claims in WPT
Opened as a working group member, not as chair

From reading the text the "ath", "tth" and "oth" claims are optional. Consider stating that as part of the claims definition. e.g. 

"- ath: Optional. Hash of the OAuth access token, if present in the request,
         which might convey end-user identity and authorization context
         of the request.  The value, as per Section 4.1 of [RFC9449], is
         the base64url encoding of the SHA-256 hash of the ASCII
         encoding of the access token's value.

yaronf:
Editorial. Closes #91.

### DIFF
--- a/draft-ietf-wimse-s2s-protocol.md
+++ b/draft-ietf-wimse-s2s-protocol.md
@@ -344,6 +344,8 @@ A WPT MUST contain the following:
      SHA-256 hash of the ASCII encoding of the associated token's value.
      (Note: this is less than ideal but seems we need something like this for extensibility.)
 
+To clarify: the `ath`, `tth` and `oth` claims are each mandatory if the respective token is included in the request.
+
 An example WPT might look like the following:
 
 ~~~ jwt


draft-ietf-wimse-s2s-protocol.md
```diff
@@ -344,6 +344,8 @@ A WPT MUST contain the following:
      SHA-256 hash of the ASCII encoding of the associated token's value.
      (Note: this is less than ideal but seems we need something like this for extensibility.)
 
+To clarify: the `ath`, `tth` and `oth` claims are each mandatory if the respective token is included in the request.
+
```
jsalowey, line 348: do we need to define what to do if there is more than one token that can fit in the oth claim?

yaronf: We could say something like, if there are multiple tokens they should be joined with colons as the separator, but I'm not sure this should be in scope: the whole `oth` things is basically an extension point.

Also, we would need to define an ordering between the tokens.

jsalowey: Open issue #129

PR #137, opened 2025-07-03
Granular authentication

yaronf:


draft-ietf-wimse-s2s-protocol.md
```diff
@@ -157,6 +157,24 @@ In either case, Workload B decides whether to authorize the call.
 In certain architectures, Workload B may need to consult with an external server when making this decision.
 5. Workload B returns a response to Workload A, which may be an error response or a regular one.
 
+## Workload Identifiers and Authentication Granularity {#granular-auth}
+
+The specific format of workload identifiers (see {{I-D.ietf-wimse-arch}}) is set by local policy for each deployment,
+and this choice has several implications.
+Pre-WIMSE, many current use cases don't allow for fully granular authentication. For instance, with mutual TLS,
+there's often no clear way to map a Request URI to the SubjectAltName value in the server certificate. This means the client can only verify
```
arndt-s, line 165: ```This means the client can only verify if the certificate is valid within a trust domain, not if it's tied to a specific workload.```

The URI SAN ties a certificate to a specific workload.

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -157,6 +157,24 @@ In either case, Workload B decides whether to authorize the call.
...
+if the certificate is valid within a trust domain, not if it's tied to a specific workload.
+
+To enable mutual and granular authentication between workloads, two things must be in place:
+
+- Each workload must know its own identifier.
+- There needs to be a two-way mapping between the Request URI used to access a workload and its workload identifier.
```
arndt-s, line 171: What's the two-way here? Isn't a request URI -> workload identifier enough? 

However, there needs to be differentiation between whether it's a client or a server situation.

`/hello` -> `wimse://example.org/my-workload` 
could mean that:
- A: my-workload is authenticated to serve `/hello`
- B: my-workload is authenticated to call `/hello`

arndt-s (approved):

PR #146, opened 2025-08-06
Clarify DPoP and middle boxes

### LINKED ISSUE #70: Middleboxes and PoP mitigations
Sec. 6.3 states that "Mitigations listed in the previous section can be used to provide some protection from middle boxes." I don't think any of the mitigations is Sec. 6.2 are effective against a middlebox that replaces a part of the message not covered by PoP.

@jsalowey

yaronf:
Resolves #70.


bc-pi (approved):
no objection

PR #151, opened 2025-08-13
Additional (non-standard) claims in WIT and WPT

### LINKED ISSUE #112: Additional claims in WPT
During the SPIFFE side of WIMSE discussions we talked about having the ability of the WPT signer (the workload) to add additional claims inside the token. There's nothing preventing the workload from doing this, so to allow better interoperability of implementations it would be good if there was some prescribed way of doing this.

The main use case for this would be to further scope down the token to a specific request. For example in the case of HTTP this could take the form of `htm` and `htu` field or even a field for a kind of signature. The WPT could then be treated as a sort of protocol agnostic token that can be adapted to various protocols. We could even specify standard ways of doing it for things such as HTTP.

The same fields could also be used in other non-protocol specific ways. For example for a CI use case a validator could validate that the token comes from a CI worker and then from the additional claims it can determine which jobs it was running.

This is currently somewhat weakly specified by the standard through the `oth` field. A workload could include a token with the additional fields and binding by including the hash of it in `oth`. To me that seems like a roundabout way of doing it where we pay the extra price of encoding the information in this way and sending more bytes on the wire without getting any other benefits. It would be nice if we could just include the additional claims as a sub-object in a claim.

Another way of getting this extra claims is through transaction tokens. This forces the workload the leave aside the benefits of WIT/WPT, including in process signing of WPTs, and resort to talking to a 3rd party to get an extra token signed. For information that is mostly relevant to a point to point communication (e.g. I'm calling this RPC with this information), it would be much more useful if the workload could include that information in the token when creating the WPT.

yaronf:
Closes #112.

### DIFF
--- a/draft-ietf-wimse-s2s-protocol.md
+++ b/draft-ietf-wimse-s2s-protocol.md
@@ -326,10 +326,11 @@ etc., are all valid and equivalent header field names. However, case is signific
 
 ### Including Additional Claims {#add-claims}
 
-The WIT is a JSON structure and therefore can be trivially extended by adding more claims than are defined in the current specification.
+The WIT is a JSON structure and therefore can be trivially extended by adding more claims beyond those defined in the current specification.
 This, however, could result in interoperability issues, which the following rules are addressing.
 
-* Deployers may freely add claims to the WIT beyond those defined here, in closed environments. Such claims SHOULD be collision-resistant, such as `example.com/myclaim`.
+* To ensure interoperability in WIMSE environments, the use of Private claim names (Sec. 4.3 of {{RFC7519}}) is NOT RECOMMENDED.
+* In closed environments, deployers MAY freely add claims to the WIT. Such claims SHOULD be collision-resistant, such as `example.com/myclaim`.
 * A recipient that does not understand such claims MUST ignore them, as per Sec. 4 of {{RFC7519}}.
 * Outside of closed environments, new claims MUST be registered with IANA {{IANA.JWT.CLAIMS}} before they can be used.
 
--- a/draft-ietf-wimse-s2s-protocol.md
+++ b/draft-ietf-wimse-s2s-protocol.md
@@ -324,6 +324,17 @@ Note that per {{RFC9110}}, header field names are case insensitive;
 thus, `Workload-Identity-Token`, `workload-identity-token`, `WORKLOAD-IDENTITY-TOKEN`,
 etc., are all valid and equivalent header field names. However, case is significant in the header field value.
 
+### Including Additional Claims {#add-claims}
+
+The WIT is a JSON structure and therefore can be trivially extended by adding more claims than are defined in the current specification.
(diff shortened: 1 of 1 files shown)


draft-ietf-wimse-s2s-protocol.md
```diff
@@ -324,6 +324,15 @@ Note that per {{RFC9110}}, header field names are case insensitive;
...
 etc., are all valid and equivalent header field names. However, case is significant in the header field value.
 
+### Including Additional Claims {#add-claims}
+
+The WIT is a JSON structure and therefore can be trivially extended by adding more claims than are defined in the current specification.
+This, however, could result in interoperability issues, which the following rules are addressing.
+
+* Deployers may freely add claims to the WIT beyond those defined here, in closed environments. Such claims SHOULD be collision-resistant, such as `example.com/myclaim`.
+* A recipient that does not understand such claims MUST ignore them, as per Sec. 4 of {{RFC7519}}.
+* Outside of closed environments, new claims MUST be registered with IANA {{IANA.JWT.CLAIMS}} before they can be used.
```
bc-pi, line 334: seems like these rules are not aligned with similar content from RFC 7519 itself, which doesn't seem good

yaronf: @bc-pi In what way?

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -324,6 +324,15 @@ Note that per {{RFC9110}}, header field names are case insensitive;
 thus, `Workload-Identity-Token`, `workload-identity-token`, `WORKLOAD-IDENTITY-TOKEN`,
 etc., are all valid and equivalent header field names. However, case is significant in the header field value.
 
+### Including Additional Claims {#add-claims}
+
+The WIT is a JSON structure and therefore can be trivially extended by adding more claims than are defined in the current specification.
```
bc-pi, line 329: the WIT is not a JSON structure. it contains some JSON but is not JSON

bc-pi: github says this is outdated but who is it to day what's in style or not?

draft-ietf-wimse-s2s-protocol.md
```diff
@@ -324,6 +324,16 @@ Note that per {{RFC9110}}, header field names are case insensitive;
 thus, `Workload-Identity-Token`, `workload-identity-token`, `WORKLOAD-IDENTITY-TOKEN`,
 etc., are all valid and equivalent header field names. However, case is significant in the header field value.
 
+### Including Additional Claims {#add-claims}
+
+The WIT is a JSON structure and therefore can be trivially extended by adding more claims beyond those defined in the current specification.
```
bc-pi, line 329:
```suggestion
The WIT contains JSON structures and therefore can be trivially extended by adding more claims beyond those defined in the current specification.
```

bc-pi (approved):

Issue #139, opened 2025-07-14
better [media] typ[e] values?


bc-pi:
wimse-id+jwt and wimse-proof+jwt could be just wit+jwt and wpt+jwt

bc-pi:
though marked as `small`, it doesn't seem small to do with #147 outstanding

bc-pi:
#152

PR #152, opened 2025-08-25
update the [media]typ[e] values

bc-pi:
for #139

change wimse-id+jwt and wimse-proof+jwt to wit+jwt and wpt+jwt respectively 
 
update examples/main.go and try to make things that don't need to change not change

updated the typ value in WITs of http sigs


draft-ietf-wimse-s2s-protocol.md
```diff
@@ -500,7 +500,8 @@ Implementors need to be aware that the WIT is extracted from the message before
 Either client or server MAY send an `Accept-Signature` header, but is not required to do so. When this header is sent, it MUST include the header components listed above.
 
 Following is a non-normative example of a signed request and a signed response,
-where the caller is using the keys specified in {{example-caller-jwk}}.
+where the caller is using the keys specified in {{example-caller-jwk}}
+(it is actually using a different key but that'll need to be fixed later).
```
yaronf, line 504:
```suggestion
(TODO: it is actually using a different key but that'll need to be fixed later).
```

yaronf (approved):

PR #179, opened 2025-10-22
draft-ietf-wimse-http-signature - let's go!

yaronf:
Preview: https://ietf-wg-wimse.github.io/draft-ietf-wimse-s2s-protocol/ys-http-sig-00/draft-ietf-wimse-http-signature.html


jsalowey (approved):
This looks fine as a draft -00,  You can reject my text suggestion if you do not think its appropriate.

PR #193, opened 2025-12-16
Delete old s2s document

arndt-s:


draft-ietf-wimse-s2s-protocol.md
arndt-s: @arndt-s  update readme

bc-pi (approved):

Issue #221, opened 2026-03-22
"aud" in HTTP-Sig - express as a derived component, not an HTTP header


yaronf:
Main reason: @bc-pi is on YouTube saying it's ugly :-)

The HTTP header was basically a misunderstanding of what an RFC 9421 derived component is allowed to be. The `aud` value actually meets the requirements, and so we can drop the HTTP header and add a derived component named `wimse-aud` to do the work.

yaronf:
However, the whole point of `aud` is that we need to carry the value explicitly, because the request URI may be rewritten. So instead of a derived component, we could use a signature parameter. Not terribly elegant but workable.

yaronf:
The PR keeps this issue open, because examples in the draft have not been updated. These examples would require some serious code changes if we agree on this approach.

yaronf:
Closed by #222.

PR #222, opened 2026-03-22
HTTP-Sig: wimse-aud signature parameter (#221)

yaronf:
## Summary
Replaces the `Wimse-Audience` HTTP header with the `wimse-aud` signature metadata parameter (RFC 9421) and adds an IANA registration in the HTTP Signature Metadata Parameters registry. Addresses #221.

## Notes
The non-normative request example updates `Signature-Input` (adds `wimse-aud`, drops the header), but the `Signature` value was not recomputed over the new signature base; follow-up to regenerate with tooling when ready.


yaronf:
@jricher

jsalowey (approved):

draft-ietf-wimse-http-signature.md
```diff
@@ -141,14 +142,17 @@ For clarity: the signature's lifetime (the `expires` signature parameter) is dif
...
-"authority" (often, the target host).
+{{RFC9421}} defines signature parameters for HTTP message signatures: metadata carried in the `Signature-Input` field
+alongside the covered components. That metadata is covered by the signature as the `@signature-params` component value
+(Section 2.3 of {{RFC9421}}), which is always the last line of the signature base.
 
-The header field contains the HTTP target URI ({{Section 7.1 of !RFC9110}}) of the request, without query or fragment parts. However, there may be some normalization, rewriting or other processes that require the audience to be set to a deployment-specific value.
+This document defines the `wimse-aud` signature metadata parameter for requests. Using a signature parameter carries the audience explicitly in `Signature-Input`.
+
+The value is the HTTP target URI ({{Section 7.1 of !RFC9110}}) of the request, without query or fragment parts.
+However, there may be normalization, rewriting, or other processes that require the audience to be set to a
```
jricher: While I realize this is the same definition as the old header, this value feels under specified. Is it deployment specific, or is it the url? What are the requirements on the verifier to validate it, if any?

Otherwise the definition looks good and serves the purpose.

yaronf: @jricher the new text should address your comment.

PR #223, opened 2026-04-21
Another attempt at simultaneous use

### LINKED ISSUE #87: Simultaneous use of authentication methods
Opened as a working group member, not as chair

The abstract states that:

These two protocols are compatible, in the sense that a
   single call chain can have some calls use one protocol and some use
   the other.  Service A can call Service B with mutual TLS
   authentication, while the next call from Service B to Service C would
   be authenticated at the application level.

Is simultaneous use supported as well? For example I may use mutual TLS for workload-to-workload authentication by deploying a service mesh, but also do application layer authentication?

arndt-s:
Suceeds #192, resolves #87


draft-ietf-wimse-workload-creds.md
```diff
@@ -169,6 +169,26 @@ Once these conditions are met, the methods described in this document can be use
...
+│           │◄─────►│  Proxy  │◄──────►│           │
+│ Workload  │       └─────────┘        │ Workload  │
+│     A     │                          │     B     │
+│           │◄========================►│           │
+└───────────┘     Application-layer    └───────────┘
+~~~
+
+Whether simultaneous use of credentials is appropriate is governed by local policy. Deployments SHOULD define clear rules for when and how multiple credentials are used on the same hop. See {{security-simultaneous-use}} for security considerations related to simultaneous use.
+
+Using different authentication methods inconsistently between the same pair of workloads SHOULD be avoided. A given pair of communicating workloads SHOULD use a consistent set of authentication methods across their interactions to prevent confusion about the required level of authentication.
```
arndt-s: Let's cut this part.

draft-ietf-wimse-workload-creds.md
arndt-s, line 1: we need to talk about identifier here. They may and likely will differ.

draft-ietf-wimse-workload-creds.md
```diff
@@ -169,6 +169,26 @@ Once these conditions are met, the methods described in this document can be use
 Implementations MUST allow for defining this mapping between the workload's access path and the workload identifier (e.g., through
 callback functions). Deployments SHOULD use these features to establish a consistent set of identifiers within their environment.
 
+## Simultaneous Use of Credentials {#simultaneous-use}
```
arndt-s, line 171: Let's make this proxy specific and less generic.

draft-ietf-wimse-workload-creds.md
arndt-s: instead of proxy -> intermediary.

arndt-s: also check architecture document for the wording used there.

draft-ietf-wimse-workload-creds.md
```diff
@@ -169,6 +169,24 @@ Once these conditions are met, the methods described in this document can be use
...
```
arndt-s, line 171: Yaron: As a workload are you obliged to use the same identifier on both?

arndt-s: Should + justification to simplify policy management & auditing

draft-ietf-wimse-workload-creds.md
```diff
@@ -456,6 +476,12 @@ Mitigations listed in {{app-level}} can be used to provide some protection from
 
 Deployments should perform analysis on their situation to determine if it is appropriate to trust and allow traffic to pass through a middle box.
 
+## Simultaneous Use of Credentials {#security-simultaneous-use}
+
+Using both transport-level and application-level authentication between the same pair of workloads (as opposed to the intermediary pattern described in {{simultaneous-use}}) SHOULD NOT be done without careful analysis. When both methods authenticate the same hop directly, it creates ambiguity about which identity is authoritative for authorization decisions and increases the attack surface without a clear security benefit.
+
+When an intermediary is present and the transport-level and application-level credentials carry different Workload Identifiers, the authorization policy MUST account for both identities and define which is authoritative for each authorization decision. Failing to do so can allow a compromised intermediary, or a misrouted request, to be authorized on the basis of the wrong identity.
```
yaronf: This is confusing: the first paragraph is expressly NOT about intermediaries, so why does the second paragraph belong here? In fact, why not move it to the first "Simultaneous Use of Credentials " section?

BTW it is not great to have two sections with the same name. "Simultaneous use of section name."

arndt-s:
Closed in favor of https://github.com/ietf-wg-wimse/draft-ietf-wimse-arch/pull/125

PR #226, opened 2026-04-23
WIT `sub` claim (#202)

### LINKED ISSUE #202: URI should be wimse identifier (?)
https://github.com/ietf-wg-wimse/draft-ietf-wimse-s2s-protocol/blob/41fb87003b43adccd77af7e91ad06e5e3499066f/draft-ietf-wimse-workload-creds.md?plain=1#L196

Rather than just a URI, shouldn't the 'sub' claim contain a WIMSE identifier ?

yaronf:
Closes #202

### DIFF
--- a/draft-ietf-wimse-workload-creds.md
+++ b/draft-ietf-wimse-workload-creds.md
@@ -59,7 +59,6 @@ informative:
   IANA.JOSE.ALGS: IANA.jose_web-signature-encryption-algorithms
   IANA.JWT.CLAIMS: IANA.jwt_claims
   IANA.MEDIA.TYPES: IANA.media-types
-  IANA.URI.SCHEMES: IANA.uri-schemes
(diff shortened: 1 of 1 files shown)


bc-pi:
I'm honestly struggling to understand "WIT sub claim" and the changes here. Some of which seem to have been done in a different change #225.

yaronf:
> I'm honestly struggling to understand "WIT sub claim" and the changes here. Some of which seem to have been done in a different change #225.

@bc-pi This PR addresses #202, and specifically this text:

> Also, in the second sentence, I don't think it's really security implications. Rather, it's additional requirements associated with these identifiers.

So it's just a minor editorial changes. I had to pull in some of #225 to avoid merge conflicts.

bc-pi (approved):
that's different than how I understand things to work but okay

PR #235, opened 2026-06-02
Unify client authorization for WIT and WIC (#228)

### LINKED ISSUE #228: Workload credentials draft specifies fetching id from certificate
Currently the [workload credentials](https://github.com/ietf-wg-wimse/draft-ietf-wimse-s2s-protocol/blob/main/draft-ietf-wimse-workload-creds.md#client-authorization-using-the-workload-identity-client-name) draft specifies fetching the workload identity from the certificate subject alternative names. This is only available in the [m]TLS case (and it's indeed reproduced there too), but that does not rely on WITs.

The section should be updated to reference fetching it from the subject claim of the WIT or move it to the WPT and HTTP signatures one, if there are differences between the two.

yaronf:
Add a top-level section on using the workload identifier for authorization. Shared policy text; bullets distinguish application level (`sub`) from transport level (certificate SAN). Use "receiving application" for the party that authorizes the peer.

Closes #228.

### DIFF
--- a/draft-ietf-wimse-workload-creds.md
+++ b/draft-ietf-wimse-workload-creds.md
@@ -359,7 +359,7 @@ See {{granular-auth}} on additional security implications of workload identifier
 How the receiving application obtains the workload identifier depends on the credential used to authenticate the peer:
 
 * Application-level: After successfully validating the WIT, the receiving application retrieves the workload identifier from the `sub` claim.
-* Transport level: When the client is authenticated with a Workload Identity Certificate, the receiving application retrieves the workload identifier from the client certificate's SubjectAltName extension of type URI.
+* Transport-level: When the client is authenticated with a Workload Identity Certificate, the receiving application retrieves the workload identifier from the client certificate's SubjectAltName extension of type URI.
 
 # Implementation Status
 
--- a/draft-ietf-wimse-workload-creds.md
+++ b/draft-ietf-wimse-workload-creds.md
@@ -358,7 +358,7 @@ See {{granular-auth}} on additional security implications of workload identifier
 
 How the receiving application obtains the workload identifier depends on the credential used to authenticate the peer:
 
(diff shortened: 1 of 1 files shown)


arndt-s (approved):

draft-ietf-wimse-workload-creds.md
```diff
@@ -349,13 +349,18 @@ As noted in the introduction, for many deployments, transport-level protection o
...
 
-The server application retrieves the workload identifier from the client certificate subjectAltName. The identifier is used in authorization, accounting and auditing.
+The receiving application uses the authenticated peer's workload identifier in authorization, accounting, and auditing.
 For example, the full workload identifier may be matched against ACLs to authorize actions requested by the peer and the identifier may be included in log messages to associate actions to the client workload for audit purposes.
 A deployment may specify other authorization policies based on the specific details of how the workload identifier is constructed. The path portion of the workload identifier MUST always be considered in the scope of the trust domain.
 See {{granular-auth}} on additional security implications of workload identifiers.
 
+How the receiving application obtains the workload identifier depends on the credential used to authenticate the peer:
+
+* Application level: After successfully validating the WIT, the receiving application retrieves the workload identifier from the `sub` claim.
```
```suggestion
* Application-level: After successfully validating the WIT, the receiving application retrieves the workload identifier from the `sub` claim.
```

draft-ietf-wimse-workload-creds.md
```diff
@@ -349,13 +349,18 @@ As noted in the introduction, for many deployments, transport-level protection o
...
+* Transport level: When the client is authenticated with a Workload Identity Certificate, the receiving application retrieves the workload identifier from the client certificate's SubjectAltName extension of type URI.
```
```suggestion
* Transport-level: When the client is authenticated with a Workload Identity Certificate, the receiving application retrieves the workload identifier from the client certificate's SubjectAltName extension of type URI.
```

PR #236, opened 2026-06-02
Add trust anchors section; clarify WIT validation (#170)

### LINKED ISSUE #170: Create section about trust anchors
We currently don't talk about where the trust anchor comes from and what data is used to identify those. We believe this should be added on a high level. The exact provisioning is out of scope.

yaronf:
Introduce normative Establishing Trust Anchors (OOB provisioning, multiple keys, `kid`/RFC 7515/8725, WIC pointer). Align `iss` note with trust anchors and RFC 8414 only when configured OOB. Update WIT example signing-key wording.

Closes #170.

### DIFF
--- a/draft-ietf-wimse-workload-creds.md
+++ b/draft-ietf-wimse-workload-creds.md
@@ -174,6 +174,16 @@ All terminology in this document follows {{?I-D.ietf-wimse-arch}}.
 
 {::boilerplate bcp14-tagged}
 
+# Establishing Trust Anchors {#trust-anchors}
+
+This specification expects that trust anchors used to validate Workload Identity Certificates and Workload Identity Tokens are associated with a trust domain and are provisioned to consumers together with trust domain metadata for that domain. Trust anchors can be represented in different ways depending on the deployment; for example, as a set of public keys (such as a JWKS), as an X.509 certification authority used to validate certificate chains, or as other deployment-specific material.
+
+This specification does not define a mechanism for discovering trust anchor information. Consumers MUST bind each trust domain to an authorized issuer (or set of issuers) and to the corresponding trust anchors. That mapping MUST be distributed via a secure out-of-band mechanism. In particular, the `iss` claim MUST NOT be used to look up trust anchor material directly from information carried only in the token (such as by fetching a JWKS from a URL found only in the WIT).
+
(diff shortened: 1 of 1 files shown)


draft-ietf-wimse-workload-creds.md
```diff
@@ -174,6 +174,16 @@ All terminology in this document follows {{?I-D.ietf-wimse-arch}}.
 
 {::boilerplate bcp14-tagged}
 
+# Establishing Trust Anchors {#trust-anchors}
+
+This specification expects that trust anchors used to validate Workload Identity Certificates and Workload Identity Tokens are associated with a trust domain and are provisioned to consumers together with other trust domain configuration (such as which issuers are authorized). Trust anchors can be represented in different ways depending on the deployment; for example, as a set of public keys (such as a JWKS), as an X.509 certification authority used to validate certificate chains, or as other deployment-specific material. A trust anchor for WIT validation MAY include more than one public key so that operators can rotate signing keys.
```
arndt-s: > and are provisioned to consumers together with other trust domain configuration (such as which issuers are authorized).

Is this metadata about the trust domain or other trust domains? I believe it's the former but it should be more explicit. Maybe "trust domain metadata"?

arndt-s: > A trust anchor for WIT validation MAY include more than one public key so that operators can rotate signing keys.

Doesn't the same apply for X.509 where a trust anchor can have multiple CA certs?

arndt-s: Maybe moving it to the third paragraph makes it more clear that this is specific to WIT and doesn't prelude any multi-CA usage in WIC

draft-ietf-wimse-workload-creds.md
```diff
@@ -174,6 +174,16 @@ All terminology in this document follows {{?I-D.ietf-wimse-arch}}.
...
+
+To validate a WIT, the receiving workload verifies the JWS signature using key material from the trust anchors configured for the trust domain in the `sub` claim. Verification follows {{RFC7515}} and the JWT best practices in {{RFC8725}}: the `alg` header parameter MUST identify an algorithm that the validator accepts for that trust domain, and the signature MUST verify under a key drawn from the configured trust anchors. If the JWS header contains a `kid` parameter, the validator MUST use the key in the trust anchor material with the same `kid` value. If `kid` is absent, the validator MUST obtain unambiguous key selection by deployment policy (for example, when only a single key is configured for that trust domain).
```
arndt-s: Discussed in the call today and an idea is to remove the normative language and explain it as an example.

arndt-s (approved):

jsalowey (approved):

Issue #240, opened 2026-06-11
Response signing: wimse-aud gap and replay to a different client


yaronf:
I think the lack of wimse-aud on the response is a real gap, in that responses can be replayed to a different client. Two ways to address it:

* Bind to the request by including a `nonce-req` value in the response (copying the `nonce` from the request). This is somewhat hacky.
* Include `wimse-aud` in the response, but that needs to be a WIMSE ID (the caller doesn't necessarily have a request URI) and for symmetry, we would need to change it to a WIMSE ID on the request too. This would change the mapping calculus again.

yaronf:
Addressed by #253.

Issue #175, opened 2025-09-26
Back to "aud"


yaronf:
* When using mTLS, the SubjectAltName is a WIMSE ID. The assumption is that both peers know each other's ID and can validate the incoming certs.
* Sec. 1.3 is about mapping from Request URI (which is known to the client by definition) to the target's WIMSE ID.

And yet we still specified the `aud` claim of the WPT (Sec. 3.2) as the Request URI. Both points above point towards `aud` as the WIMSE ID.

yaronf:
I suggest we close this issue without changing `aud`/`wimse-aud` to a Workload Identifier:

1. **Keep audience as the HTTP target URI** in WPT and HTTP Signatures (RFC 9110 §7.1, no query/fragment, existing normalization/rewriting caveats). The caller knows this URI when constructing the request. In HTTP Signatures, `@request-target` covers path (and query); `wimse-aud` adds scheme and authority. Identity (`sub`, WIC SAN) identifies the sender; audience binds the PoP to the HTTP access handle, not a WIMSE ID.

2. **Keep the #granular-auth mapping**, direction access path → Workload Identifier (#173 is a separate editorial fix). The creds draft already requires a concrete mapping API (e.g. callback functions). Either direction would work; we keep the existing one for interoperability. The caller sets `aud`/`wimse-aud` to the request URI; the callee uses the mapping to validate audience (including alias/normalization).

3. **Which side uses the mapping depends on the method:**
   * **mTLS:** caller — server validation by workload identity (#server-name, #granular-auth). Callee gets caller ID from the client cert.
   * **WPT / HTTP Signatures:** callee — PoP audience validation.

yaronf:
Discussion: implement req/res binding in http-sig, leave wimse-aud as RequestURI, and resolve the mapping issue.

PR #246, opened 2026-06-19
HTTP target URI audience model and response request binding (#173, #175, #240)

### LINKED ISSUE #173: Unclear text about mapping of identifiers
Sec. 1.3 defines a mapping from the Request URI to the WIMSE ID, but then the MUST statement at the bottom requires a mapping "between" these two, which can be read as a bidirectional mapping.

yaronf:
- **workload-creds (#173, #175):** Replace the mandatory access-path-to-identifier mapping API with infrastructure support for verifying that an HTTP target URI (RFC 9110 Section 7.1, without query or fragment) is intended for a workload. Document the roles of HTTP target URI (audience) vs Workload Identifier (sender) for WPT, HTTP signatures, and mTLS.
- **WPT (#175):** Tie `aud` to "Workload Identifiers and Authentication Granularity" in workload-creds; otherwise unchanged from main.

Closes #173, #175. Addresses #240.

### DIFF
--- a/draft-ietf-wimse-http-signature.md
+++ b/draft-ietf-wimse-http-signature.md
@@ -158,7 +158,7 @@ The default value for `wimse-aud` is the request's HTTP target URI ({{Section 7.
 Senders, recipients, and intermediaries do not always derive the same string for that URI: normalization and rewriting differ by implementation and hop, so the audience that verification should use is a deployment-specific choice.
 When the default string is not suitable for verification at the recipient, senders SHOULD set `wimse-aud` to an explicit audience value as appropriate for that deployment.
 
-The recipient MUST be able to verify that the audience is intended for it, using trusted configuration or a local API (for example, to determine the expected request URI or to validate a received audience value); see "Workload Identifiers and Authentication Granularity" in {{I-D.ietf-wimse-workload-creds}}.
+The recipient MUST be able to verify that the audience refers to it. See "Workload Identifiers and Authentication Granularity" in {{I-D.ietf-wimse-workload-creds}} for more detail.
 
 ## The `wimse-req-nonce` Signature Parameter {#wimse-req-nonce-param}
 
--- a/draft-ietf-wimse-http-signature.md
+++ b/draft-ietf-wimse-http-signature.md
@@ -394,7 +394,7 @@ IANA is requested to register the following entries in the "HTTP Signature Metad
 
 ## draft-ietf-wimse-http-signature-04
 
-* On signed responses, require `wimse-req-nonce` (request binding); register with IANA.
+* On signed responses, require `wimse-req-nonce` (request binding); register with IANA. Non-normative response example not updated accordingly; the `Signature` value was not regenerated (see issue tracker).
 
 ## draft-ietf-wimse-http-signature-03
 
--- a/draft-ietf-wimse-http-signature.md
+++ b/draft-ietf-wimse-http-signature.md
@@ -125,6 +125,10 @@ For requests only, the following signature parameter MUST also be included:
 
 * `wimse-aud` ({{wimse-aud-param}})
 
+For responses only, when response signing is enabled, the following signature parameter MUST also be included:
+
+* `wimse-req-nonce` ({{wimse-req-nonce-param}})
+
 The following signature parameters in the `Signature-Input` header MUST NOT be used:
 
 * `keyid` - The signing key is sent along with the message in the WIT. Additionally specifying the key identity would add confusion.
@@ -154,7 +158,15 @@ The default value for `wimse-aud` is the request's HTTP target URI ({{Section 7.
 Senders, recipients, and intermediaries do not always derive the same string for that URI: normalization and rewriting differ by implementation and hop, so the audience that verification should use is a deployment-specific choice.
 When the default string is not suitable for verification at the recipient, senders SHOULD set `wimse-aud` to an explicit audience value as appropriate for that deployment.
 
-The recipient MUST be able to verify that the audience refers to it. See "Workload Identifiers and Authentication Granularity" in {{I-D.ietf-wimse-workload-creds}} for more detail.
+The recipient MUST be able to verify that the audience is intended for it, using trusted configuration or a local API (for example, to determine the expected request URI or to validate a received audience value); see "Workload Identifiers and Authentication Granularity" in {{I-D.ietf-wimse-workload-creds}}.
+
+## The `wimse-req-nonce` Signature Parameter {#wimse-req-nonce-param}
+
+When response signing is enabled, this document defines the `wimse-req-nonce` signature metadata parameter for responses.
+This parameter binds requests to responses and prevents a malicious
+proxy from replaying responses to the wrong client.
+
+The server MUST set `wimse-req-nonce` to the value of the `nonce` signature parameter from the `Signature-Input` of the request that triggered the response.
 
 ## Signing the Response
 
@@ -164,6 +176,8 @@ may be exceptionally large or is expected to be streamed, signing it may not be 
 In practice, we expect response signing to be enabled by local policy. If response signing is enabled for a deployment,
 the client (recipient of the response) MUST check that the signature exists and validate it.
 The response MUST be rejected if a signature is absent or fails to validate.
+The client MUST verify that `wimse-req-nonce` matches the `nonce` it included in its request's `Signature-Input`.
+This binds the signed response to the specific request that triggered it.
 
 As described in {{Section 5 of RFC9421}}, either client or server MAY send an
 `Accept-Signature` header,
@@ -333,7 +347,7 @@ been activated and likewise, the recipient validates this signature.
 * No requests can be modified without detection by the recipient. Integrity of
   all present HTTP headers specified in this document is protected, as well as
 the derived components listed in {{http-sig-auth}}, the signature parameters
-(including `wimse-aud` on requests) as covered by `@signature-params` in {{RFC9421}}, and
+(including `wimse-aud` on requests and `wimse-req-nonce` on responses) as covered by `@signature-params` in {{RFC9421}}, and
 the message body (when present).
 * No responses can be modified without detection, provided that optional response signing has been activated and
 that the recipient validates incoming responses.
@@ -348,6 +362,7 @@ caches across validators). Therefore it is not claimed as
 a goal, though implementations SHOULD attempt to detect replays where feasible.
 We note that since most of the message is signed, replay attacks are only possible in a
 context where the request would be accepted as valid, and this mitigates the risk to some extent.
+* When response signing is enabled, validating `wimse-req-nonce` mitigates replay of a signed response to a client other than the one that sent the triggering request.
 * Unless response signing is mandated by local policy, complete deletion of a request/response pair is possible without detection.
 
 
@@ -355,9 +370,10 @@ context where the request would be accepted as valid, and this mitigates the ris
 
 ## HTTP Signature Metadata Parameters Registration
 
-IANA is requested to register the following entry in the "HTTP Signature Metadata Parameters" registry {{IANA.HTTP.MESSAGE.SIGNATURE}}, per the registration template in Section 6.3.1 of {{RFC9421}}:
+IANA is requested to register the following entries in the "HTTP Signature Metadata Parameters" registry {{IANA.HTTP.MESSAGE.SIGNATURE}}, per the registration template in Section 6.3.1 of {{RFC9421}}:
 
 * `wimse-aud`, per {{iana-wimse-aud-param}}.
+* `wimse-req-nonce`, per {{iana-wimse-req-nonce-param}}.
 
 ### `wimse-aud` {#iana-wimse-aud-param}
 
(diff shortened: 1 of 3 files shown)


draft-ietf-wimse-workload-creds.md
```diff
@@ -160,13 +160,11 @@ if the server certificate is valid within a trust domain, not if it's tied to a
 To enable mutual and granular authentication between workloads, two things must be in place:
 
 - Each workload must know its own identifier.
-- There needs to be an explicit mapping from the external handle used to access a workload (such as an Ingress path or service DNS name)
-to its workload identifier.
+- The infrastructure must make available to workloads some means to verify that an HTTP target URI ({{Section 7.1 of RFC9110}}), without query or fragment components, is intended for them.
```
yaroslavros:
```suggestion
- Each workload must be able to verify that an HTTP target URI ({{Section 7.1 of RFC9110}}) is intended for it.
```

draft-ietf-wimse-workload-creds.md
```diff
@@ -160,13 +160,11 @@ if the server certificate is valid within a trust domain, not if it's tied to a
...
 
-Once these conditions are met, the methods described in this document can be used for the caller and callee to mutually authenticate.
+Once these conditions are met, the methods described in this document can be used for the caller and callee to mutually authenticate. Specifically, deployments should use this capability for consistent audience validation within their environment.
```
bc-pi: mutually is too much 

```suggestion
Once these conditions are met, the methods described in this document can be used for the caller and callee to authenticate. Specifically, deployments should use this capability for consistent audience validation within their environment.
```

draft-ietf-wimse-workload-creds.md
```diff
@@ -160,13 +160,11 @@ if the server certificate is valid within a trust domain, not if it's tied to a
...
 
-Implementations MUST allow for defining this mapping between the workload's access path and the workload identifier (e.g., through
-callback functions). Deployments SHOULD use these features to establish a consistent set of identifiers within their environment.
+Application-level proof-of-possession mechanisms in {{?I-D.ietf-wimse-wpt}} and {{?I-D.ietf-wimse-http-signature}} use the HTTP target URI ({{Section 7.1 of RFC9110}}) of the request, without query or fragment components, as audience. The Workload Identifier in a WIT or WIC identifies the sender; the audience identifies the intended recipient. The caller sets the audience to the HTTP target URI; the callee verifies that the audience is intended for it using infrastructure support as described above, including any deployment-specific aliases or normalization. For mutual TLS, the caller uses the same support when validating the server by workload identity rather than DNS hostname (see {{?I-D.ietf-wimse-mutual-tls}}).
```
yaroslavros:
```suggestion
```

draft-ietf-wimse-workload-creds.md
jsalowey: I don't understand why this criteria is needed here.  I think the criteria should be.  It needs to have the information to properly validate a workload credential. 
```suggestion
- A workload must have the information needed to properly validate a workload credential
```

draft-ietf-wimse-workload-creds.md
```diff
@@ -160,13 +160,11 @@ if the server certificate is valid within a trust domain, not if it's tied to a
 To enable mutual and granular authentication between workloads, two things must be in place:
 
 - Each workload must know its own identifier.
-- There needs to be an explicit mapping from the external handle used to access a workload (such as an Ingress path or service DNS name)
-to its workload identifier.
+- The infrastructure must make available to workloads some means to verify that an HTTP target URI ({{Section 7.1 of RFC9110}}), without query or fragment components, is intended for them.
 
-Once these conditions are met, the methods described in this document can be used for the caller and callee to mutually authenticate.
+Once these conditions are met, the methods described in this document can be used for the caller and callee to authenticate. Specifically, deployments should use this capability for consistent audience validation within their environment.
```
jsalowey: This needs to be broken out into a a statement that is general for workload credentials

```suggestion
Once these conditions are met, the methods described in this document can be used for the caller and callee to authenticate one another. 
```
Then there can be discussion that is specific to the application layer credential (WIT) and the transport layer credential (x.509)

draft-ietf-wimse-workload-creds.md
jsalowey: I think this text could be OK if it is called out as specific to WIT and the TLS text is in a separate paragraph/subsection.

jsalowey (changes requested):
I find the granular authentication section confusing.  It doesn't seem like this is the right place to talk about audience.  Perhaps the audience discussion should be in a subsection about the WIT application level usage.

draft-ietf-wimse-workload-creds.md
yaronf: The intent here is for the callee to validate that a request is intended for it, in other words that it is the intended audience.

@yaroslavros although the caller (usually) sets the `aud` to the Request URI, this value might be modified in transit by proxies. So we cannot assume it is seen by the recipient.

@jsalowey this is not about validating the credential but rather about ensuring that it's targeted at this particular recipient.

draft-ietf-wimse-workload-creds.md
```diff
@@ -160,13 +160,37 @@ if the server certificate is valid within a trust domain, not if it's tied to a
...
+
+### Application-Layer Audience {#app-audience}
+
+Application-level proof-of-possession mechanisms in {{?I-D.ietf-wimse-wpt}} and {{?I-D.ietf-wimse-http-signature}}
+bind a request to an intended recipient using an audience value.
+The Workload Identifier in a WIT identifies the sender; the audience identifies the intended recipient.
+
+By default, the caller sets the audience to the HTTP target URI ({{Section 7.1 of RFC9110}}) of the request,
+without query or fragment components.
+Intermediaries may rewrite that URI in transit, so the callee MUST NOT rely solely on the request URI as
```
arndt-s, line 178: I think it's worth to clarify that the URI of the request is re-written, not the URI of the audience.
```suggestion
Intermediaries may rewrite the URI of the request in transit, so the callee MUST NOT rely solely on the request URI as
```

draft-ietf-wimse-workload-creds.md
```diff
@@ -160,13 +160,37 @@ if the server certificate is valid within a trust domain, not if it's tied to a
...
+observed on the wire to decide whether it is the intended recipient.
+Instead, the callee verifies that the audience value carried in the proof is intended for it.
+In some cases this might necessitate infrastructure support to account for deployment-specific aliases or normalization.
+Deployments SHOULD use this capability for consistent audience validation within their environment.
```
arndt-s, line 182: It's critical for the security aspects of this draft, shouldn't this be a MUST?

draft-ietf-wimse-workload-creds.md
```diff
@@ -160,13 +160,37 @@ if the server certificate is valid within a trust domain, not if it's tied to a
...
+
+### Transport-Layer Recipient Check {#transport-identity}
+
+At the transport layer there is no separate audience claim analogous to application-layer proof of possession.
```
arndt-s, line 186: Does it need HTTP in this section? What if TLS is used with a different application protocol? I suggest keeping it to TLS only?

draft-ietf-wimse-workload-creds.md
```diff
@@ -160,13 +160,37 @@ if the server certificate is valid within a trust domain, not if it's tied to a
...
 
-Implementations MUST allow for defining this mapping between the workload's access path and the workload identifier (e.g., through
-callback functions). Deployments SHOULD use these features to establish a consistent set of identifiers within their environment.
+How that recipient check is expressed depends on whether authentication is performed at the application layer or at the transport layer.
+
+### Application-Layer Audience {#app-audience}
+
+Application-level proof-of-possession mechanisms in {{?I-D.ietf-wimse-wpt}} and {{?I-D.ietf-wimse-http-signature}}
+bind a request to an intended recipient using an audience value.
+The Workload Identifier in a WIT identifies the sender; the audience identifies the intended recipient.
```
arndt-s, line 174:
```suggestion
The Workload Identifier in a WIT identifies the sender; the audience of the PoP identifies the intended recipient.
```

draft-ietf-wimse-workload-creds.md
```diff
@@ -160,13 +160,37 @@ if the server certificate is valid within a trust domain, not if it's tied to a
...
+
+By default, the caller sets the audience to the HTTP target URI ({{Section 7.1 of RFC9110}}) of the request,
+without query or fragment components.
+Intermediaries may rewrite that URI in transit, so the callee MUST NOT rely solely on the request URI as
+observed on the wire to decide whether it is the intended recipient.
+Instead, the callee verifies that the audience value carried in the proof is intended for it.
+In some cases this might necessitate infrastructure support to account for deployment-specific aliases or normalization.
+Deployments SHOULD use this capability for consistent audience validation within their environment.
+
+### Transport-Layer Recipient Check {#transport-identity}
```
yaronf, line 184: 1. It's not strictly "transport". 2. Consider removing this subsection altogether.

draft-ietf-wimse-workload-creds.md
```diff
@@ -160,13 +160,37 @@ if the server certificate is valid within a trust domain, not if it's tied to a
...
+
+At the transport layer there is no separate audience claim analogous to application-layer proof of possession.
+HTTP already requires that, once a server has reconstructed the request's target URI, it decide whether it is
+configured to process that request and whether the connection context is appropriate for it, and an origin server
+MUST reject requests that appear to have been misdirected ({{Section 7.4 of RFC9110}}).
+When connections are reused across origins — as HTTP/2 permits — a server that is not authoritative for a request
+indicates that by responding with 421 (Misdirected Request) ({{Section 9.1.1 of ?RFC9113}}).
+That check — together with ordinary TLS server authentication (including the rules in {{?I-D.ietf-wimse-mutual-tls}}) —
+is the recipient validation available when relying on mutual TLS alone.
```
bc-pi, line 193: as discussed on 8/25 
```suggestion
```

draft-ietf-wimse-http-signature.md
```diff
@@ -125,6 +125,10 @@ For requests only, the following signature parameter MUST also be included:
 
 * `wimse-aud` ({{wimse-aud-param}})
 
+For responses only, when response signing is enabled, the following signature parameter MUST also be included:
```
yaronf, line 128: Remove http-sig parts of this PR.

draft-ietf-wimse-workload-creds.md
```diff
@@ -573,6 +586,10 @@ IANA is requested to register the following entries to the "Hypertext Transfer P
 # Document History
 <cref>RFC Editor: please remove before publication.</cref>
 
+## draft-ietf-wimse-workload-creds-04
```
yaronf, line 589: Needs to be -03.

draft-ietf-wimse-workload-creds.md
```diff
@@ -161,13 +161,26 @@ but not that it belongs to the particular workload the client intended to reach.
 To enable mutual and granular authentication between workloads, two things must be in place:
 
 - Each workload must know its own identifier.
-- There needs to be an explicit mapping from the external access name used to access a workload (such as an Ingress path or service DNS name)
-to its Workload Identifier.
+- Each workload must be able to verify that it is the intended recipient of a given request
+  (that is, that it is the intended audience for that request).
```
arndt-s, line 165: and workloads need to know the identifier of the intended target of a given request.

It could be DNSName SAN, but also URI SAN.

draft-ietf-wimse-workload-creds.md
```diff
@@ -573,6 +586,10 @@ IANA is requested to register the following entries to the "Hypertext Transfer P
 # Document History
 <cref>RFC Editor: please remove before publication.</cref>
 
+## draft-ietf-wimse-workload-creds-04
+
+* Replace the access-path-to-identifier mapping API with application-layer audience for WPT and HTTP signatures (#173, #175).
```
arndt-s, line 591: needs to go to draft-ietf-wimse-workload-creds-03 - it isn't published yet.

draft-ietf-wimse-workload-creds.md
yaronf: To your first point: we do not assume that the caller knows the WIMSE ID of the callee. The caller has *some* identifier of the callee (the "external access name"), and the fact that the request goes through is proof enough that this identifier is correct.

arndt-s: > To your first point: we do not assume that the caller knows the WIMSE ID of the callee. The caller has some identifier of the callee (the "external access name"), and the fact that the request goes through is proof enough that this identifier is correct.

I don't think that's written anymore in the specification though. We RECOMMEND that a workload identity certificate carries DNSSAN. If that's not the case the caller needs to know the WIMSE-ID of the callee.

See Section 6.1
> When the certificate is also used for TLS server authentication, it is RECOMMENDED to include SubjectAltName extensions of type DNSName with the appropriate DNS names for clients that validate the server by host name rather than by workload identity. The certificate MAY contain SubjectAltName extensions of other types for PKIX or local policy, but only the URI SAN above is the WIMSE workload identity.

draft-ietf-wimse-workload-creds.md
```diff
@@ -140,20 +140,33 @@ Prior to WIMSE, many containerized runtime platforms could not authenticate a pe
...
-to its Workload Identifier.
+- Each workload must be able to verify that it is the intended recipient of a given request
+  (that is, that it is the intended audience for that request).
 
 Once these conditions are met, the methods described in this document can be used for the caller and callee to mutually authenticate.
 
-Implementations MUST allow for defining this mapping between the workload's external access name and the Workload Identifier (e.g., through
-callback functions). Deployments SHOULD use these features to establish a consistent set of identifiers within their environment.
+For application-level proof of possession, that recipient check is expressed as follows.
+
```
yaroslavros: I would remove this redundant sentence introducing the upcoming subsection.

draft-ietf-wimse-workload-creds.md
```diff
@@ -140,20 +140,33 @@ Prior to WIMSE, many containerized runtime platforms could not authenticate a pe
...
+### Application-Layer Audience {#app-audience}
+
+Application-level proof-of-possession mechanisms in {{?I-D.ietf-wimse-wpt}} and {{?I-D.ietf-wimse-http-signature}}
```
yaroslavros:
```suggestion
Application-level proof-of-possession mechanisms defined in {{?I-D.ietf-wimse-wpt}} and {{?I-D.ietf-wimse-http-signature}}
```

yaroslavros (approved):

PR #252, opened 2026-07-04
Reworked Server Name Validation section

yaroslavros:


jsalowey (approved):

yaronf (changes requested):
I'm not very happy with this PR. We are essentially defining two modes of mTLS, with and without SAN. I think this is confusing and potentially less secure. Also, it's important enough that it should be discussed (at a high level) in the Architecture doc.

yaroslavros:
I don't think there is any other solution here since TLS cannot present two certificates.

We have an unfortunately choice of:

- Using WIMSE certificates only on client side and hostname certificates on server side. This would make WIMSE incompatible with current SPIFFE practices.
- Require WIMSE certificates (with or without hostname SAN) on server side. This will make WIMSE mTLS spec unusable on public endpoints as such certificates won't be issued by webPKI.
- Provide flexibility to deployments until there is a way to provide two identity certificates in TLS

Am I missing some other option? Which path would you prefer?

yaronf:
> I don't think there is any other solution here since TLS cannot present two certificates.
> 
> We have an unfortunately choice of:
> 
> * Using WIMSE certificates only on client side and hostname certificates on server side. This would make WIMSE incompatible with current SPIFFE practices.
> * Require WIMSE certificates (with or without hostname SAN) on server side. This will make WIMSE mTLS spec unusable on public endpoints as such certificates won't be issued by webPKI.
> * Provide flexibility to deployments until there is a way to provide two identity certificates in TLS
> 
> Am I missing some other option? Which path would you prefer?

Thanks for clarifying the problem. As we all know, "flexibility" is where security issues creep in.

The question is, is it even realistic to expect public endpoints to use WIMSE/mTLS, because in real-life enterprise deployments you'd have TLS middleboxes on both ingress and egress. Once we abandon this goal, you can use WIMSE certs issued by internal CAs.

yaroslavros:
> The question is, is it even realistic to expect public endpoints to use WIMSE/mTLS, because in real-life enterprise deployments you'd have TLS middleboxes on both ingress and egress. Once we abandon this goal, you can use WIMSE certs issued by internal CAs.

Yes, it is. Dedicated public mTLS API endpoint option is becoming quite common - for example https://help.openai.com/en/articles/10876024-openai-mutual-tls-beta-program. WIMSE certificate on client side would work well for this, while server presents webPKI hostname cert.

yaronf (approved):

Issue #257, opened 2026-07-14
Section 5.1 / Figure 4 - the example uses an "iat" claim that the document never defines.


PieterKas:
The example WIT (Figures 2 and 4) contains an iat claim, but iat is not listed or defined in the Section 5.1 normative claim list (iss, sub, exp, jti, cnf). The primary example uses a claim the specification does not mention. Either add iat to the claim definitions (it is conventional and underpins the "refreshed regularly" / lifetime language) or remove it from the example. My preference would be to add it as an OPTIONAL claim.

n2ctech:
One consideration for RECOMMENDED rather than OPTIONAL: with iat present, a validator can cap effective lifetime via exp minus iat, and an auditor can place a token before or after a key rotation. An optional claim gives neither reliably. Removing iat from the examples also works. The current state, where the flagship examples carry a claim the spec never defines, is the worst of both.

(Also raised as item 5 in my WGLC review: https://mailarchive.ietf.org/arch/msg/wimse/S5Zjmr4qneFsAmepgRWhkWtFdPA/)

bc-pi:
FWIW Section 5.1 has "WIT MUST contain the following content, except where noted:" and JWTs allow for other claims unless explicitly prohibited by an application or profile.  So pedantically, this isn't a problem. That said, it might be worthwhile to add mention or guidance of such in general and/or specifically for `iat`.

bc-pi:
The the best of my recollection, which admittedly isn't very good, `iat` was originally in the list of "WIT MUST contain the following content" and therefor required. I need to look back and see if that's true, and if so, why it changed. But barring some great revelation from that, do folks see any problem with making `iat` required?

arndt-s:
What benefit would it bring to make `iat` required? Like apart from that we'd be consistent with the example? I see deployment-specific benefits but I am unsure if there's interoperability benefits that directly support a MUST.

bc-pi:
> The the best of my recollection, which admittedly isn't very good, `iat` was originally in the list of "WIT MUST contain the following content" and therefor required. 

Looking back over some of the history, it seems I was wrong about that. Apologies.

bc-pi:
> What benefit would it bring to make `iat` required? Like apart from that we'd be consistent with the example? I see deployment-specific benefits but I am unsure if there's interoperability benefits that directly support a MUST.

[...]

Issue #256, opened 2026-07-14
Section 9.2 - inconsistency in use of WIT and MTLS.


PieterKas:
The following sentence: "The WIT and its PoP are only used in the application-layer options, and both are not used in MTLS." appears to contradict Section 9.2.5 (which allows binding the PoP to a TLS session's client identity or channel-binding parameters, i.e., mTLS) and Section 1 ("This does not preclude the use of the WIT in transport-layer protocols..."). Consider deleting it to be consistent.

Issue #255, opened 2026-07-14
Section 4 - Presence of additional identifiers in WITs


PieterKas:
I agree that a credential should unambiguously identify the primary subject it was issued to, and the access decision keys off that one identity (sub for a WIT, single URI SAN for a WIC). 

I think that is in part why  the following sentence prohibit the inclusion of additional workload identifiers for correlation or any other purposes in a WIT.

"Additional correlation for logging or operations MUST NOT be encoded as a second workload identifier in the same WIT or WIC."

However, the draft allows this for certificates. Section 6.1 requires the workload identifier in exactly one URI SAN, yet explicitly allows the certificate to carry other identifiers (DNSName SANs and "SubjectAltName extensions of other types"), stating that "only the URI SAN above is the WIMSE workload identity." So a WIC can already carry multiple identifiers of different kinds while exactly one uniquely identifies the workload. The Section 4 prohibition for the WIT is inconsistent with Section 6.1.

Question: Is the intent to preclude all additional identifiers, or only to guarantee that the primary subject is unambiguous? If the latter, is the MUST NOT really needed or should it be. relaxed so that additional identifiers MAY be included for purposes such as correlation, provided they are clearly distinguished from, and never usable in place of, the authoritative subject (e.g. the WIT could include other claims that contain identifiers, but only the sub claim is treated as the primary identifier)?

arndt-s:
@PieterKas the way I see the document currently written is that "only one workload identifier is permitted per credential". This does not exclude other identifiers such as DNSName because they are not Workload Identifiers as defined by WIMSE. Same applies for WITs, jti and other correlating identifiers are totally valid.

I believe that the draft is in this consistent between WIT and WIC. Both say that additional identifiers MUST NOT be encoded as workload identifiers.

Now, do you believe additional workload identifiers as defined by WIMSE should be allowed or do you agree with our assessment?

bc-pi:
Generally concur with @arndt-s's perspective. The differences are largely due to differences in the underling structure of the credential and common usage thereof. 

Also server side credentials are kind of a special case. Even if you accept the idea of server side WIMSE credentials, and I'm not sure I do, you probably still need a DNSName SAN for fallback or coexistence with regular old TLS/HTTPS.

PieterKas:
I think some text to clarify this nuanced view would be a good thing (it did not immediately occur to me to read it this way). 

[...]

PR #267, opened 2026-07-15
Editorial fixes for workload-creds (#258–#266)

### LINKED ISSUE #258: Section 11.1.1 - broken cross-reference markup.
In the Media Type registration, "Encoding considerations... See [RFC7519]." uses markdown-link reference syntax instead of the kramdown-rfc {{RFC7519}} used everywhere else, so it will not render as a proper cross-reference. Consider changing to {{RFC7519}}.

### LINKED ISSUE #259: Section 1 -- proof-of-possession sentence.
Consider adding a sentence to the following paragraph to make it clear that authentication is done through a proof of possession mechanism:

The various protocol bindings that use these credentials to authenticate workloads to each other are specified in separate documents and are out of scope for this document. At the time of writing, three such protocol bindings are defined:

For example:

The various protocol bindings that use these credentials to authenticate workloads to each other are specified in separate documents and are out of scope for this document. Each such binding defines how a workload proves possession of the key material bound to its credential in order to authenticate to another workload. At the time of writing, three such protocol bindings are defined:

### LINKED ISSUE #260: Section 1.2 - "regular one"
The term "regular one" coule be more precise. Consider replacing:

"5. Workload B returns a response to Workload A, which may be an error response or a regular one."

With:

"5. Workload B returns a response to Workload A, which may be either a success or an error response."

### LINKED ISSUE #261: Section 1.3 - "fully granular authentication" is vague, and three synonyms are used for one concept.
The opening paragraph never pins down what granularity is meant. Consider alternate text such as:

"Prior to WIMSE, many containerized runtime platforms could not authenticate a peer as a specific workload, they could only establish that the peer belonged to a given trust domain. With mutual TLS, for example, there is often no reliable way to map the external access name a client uses to reach a workload (such as a Kubernetes Ingress path, service name, or Host header) to the SubjectAltName in the presented certificate. As a result, the client can verify that the server certificate is valid within a trust domain, but not that it belongs to the particular workload the client intended to reach."

Separately, Section 1.3 uses three undefined synonyms for the same concept: "external access reference", "external handle", and "access path". Consider using a a single term (e.g. "external access name" or "external access reference").

### LINKED ISSUE #262: Section 3 - iss claim: missing antecedent and unclear scope.
The sentence "the iss claim MUST NOT be used to look up trust anchor material..." appears in Section 3, but iss is not introduced/defined until Section 5.1 (forward reference). Consider adding a forward reference (e.g. "as defined in Section 5.1").

In addition, Section 3 covers both the WIT and the WIC, yet only the WIT has an iss claim (the WIC has an X.509 Issuer field, which is different from the iss). Consider scoping the sentence to the WIT and identify the claim on first use. For example:

"For a WIT the iss claim (the JWT issuer claim [RFC7519], defined in Section 5.1) MUST NOT be used to look up trust anchor material directly from information carried only in the token (such as by fetching a JWKS from a URL found only in the WIT)."

### LINKED ISSUE #263: Section 9.2 - "as described in Section 5.1".
Consider changing "… as described in Section 5.1" to "... as required in Section 5.1".

### LINKED ISSUE #264: Section 5.1 / Figure 4 -- the primary example omits the RECOMMENDED iss claim.
Consider either including it in the primary example or adding a one-line note that the omission is deliberate.

### LINKED ISSUE #265: Editorial consistency
Some inconcsistency in terminology that the editors might consider cleaning up:

- "specification" vs "document": pick one.
- Casing: "MTLS" (Section 9.2) vs "mutual TLS" (Section 1.3) vs "mTLS" (history). Standardize (suggest "mutual TLS" on first use, "mTLS" thereafter). "POP" (Section 9.2.5) vs "PoP" elsewhere: use "PoP".
- "middle box" (Section 9.5) vs "middlebox"/"intermediary": standardize.
- "e.g." without a following comma (Section 5.1) vs "e.g.," / "for example" elsewhere: pick one style.
- Quoting of schemes/media types: wimse (backtick) vs "spiffe" (quotes); wit+jwt (backtick) vs "application/jwt" (quotes) vs bare application/wit+jwt. Apply one convention (quotes are nice as they survive formatting/post processing)
- "Sec. 4.3" / "Sec. 4" (Section 5.1.2) vs "Section" spelled out elsewhere.
- "http auth mechanisms" (Section 5.2): Should HTTP be capitalised and auth expanded to authenticate for clarity?
- "proof of possession based WITs" (Section 5.3): consider "proof-of-possession-based".
- "host header" (Section 1.3): Did you mean the HTTP Host header field here
- Contractions ("there's", "it's") inconsistently used.

### LINKED ISSUE #266: Markup nits (kramdown source).
- Section 5.1: cnf and jwk appear without backticks in two places ("Within the cnf claim...", "Within the jwk object...") though backticked elsewhere. Consider aligning with one convention throughout.
- ABNF block: opening fence "~~~" but closing fence "~~~~" (mismatched length).
- Implementation Status: no blank line after `</cref>`, so the RFC-Editor note merges into the following paragraph (renders as a malformed "//" comment block in the .txt).

yaronf:
## Summary
Editorial-only updates to `draft-ietf-wimse-workload-creds` based on Pieter's recent issues (keeps backticks for media types/schemes).

Fixes #258
Fixes #259
Fixes #260
Fixes #261
Fixes #262
Fixes #263
Fixes #264
Fixes #265
Fixes #266

Out of scope (not editorial / deferred): #255, #256, #257.

### DIFF
--- a/draft-ietf-wimse-workload-creds.md
+++ b/draft-ietf-wimse-workload-creds.md
@@ -428,7 +428,7 @@ Teleport - Machine & Workload Identity
 
 Workload Identifiers ({{WIMSE-ID}}) are scoped to a trust domain (the URI authority component) and MUST be interpreted in that trust domain context. Using a Workload Identifier without taking into account the trust domain could allow one domain to issue tokens to spoof identities in another domain. See {{trust-anchors}} for binding trust domains to issuers and trust anchors.
 
-When a deployment uses the `iss` claim for key distribution as required in {{wit-iss-note}}, validators MUST enforce an allowlist of accepted issuers. Absent such a restriction, any entity could stand up an issuer, present a WIT with that issuer's `iss` value, and have it accepted by a validator that fetches and trusts the corresponding key material without verifying the issuer's legitimacy.
+When a deployment uses the `iss` claim for key distribution as described in {{wit-iss-note}}, validators MUST enforce an allowlist of accepted issuers. Absent such a restriction, any entity could stand up an issuer, present a WIT with that issuer's `iss` value, and have it accepted by a validator that fetches and trusts the corresponding key material without verifying the issuer's legitimacy.
 
 ## Workload Identity Token and Proof of Possession {#wit-pop}
 
--- a/draft-ietf-wimse-workload-creds.md
+++ b/draft-ietf-wimse-workload-creds.md
@@ -336,7 +336,9 @@ This, however, could result in interoperability issues, which the following rule
 
 ### A note on `iss` claim and key distribution {#wit-iss-note}
 
-It is RECOMMENDED that the WIT carries an `iss` claim, including for the auditing and operational uses described above. Validators are not required to use `iss` when validating the WIT or establishing the workload identity: the trust domain and workload identity are carried in the mandatory `sub` claim ({{WIMSE-ID}}). Implementations MAY include the `iss` claim in the form of a `https` URL to facilitate key distribution via mechanisms like the `jwks_uri` from {{!RFC8414}}, only when the issuer and verification keys (or metadata such as `jwks_uri`) are configured out of band as in {{trust-anchors}}. Alternative key distribution methods may use only the trust domain from the `sub` claim.
+It is RECOMMENDED that the WIT carries an `iss` claim for the auditing and operational uses described above. Validators are not required to use `iss` when validating the WIT or establishing the workload identity: the trust domain and workload identity are carried in the mandatory `sub` claim ({{WIMSE-ID}}).
+
+Implementations MAY include the `iss` claim in the form of a `https` URL to facilitate key distribution via mechanisms like the `jwks_uri` from {{!RFC8414}}, only when the issuer and verification keys (or metadata such as `jwks_uri`) are configured out of band as in {{trust-anchors}}. Alternative key distribution methods may use only the trust domain from the `sub` claim.
 
 ## Error Conditions
 
--- a/draft-ietf-wimse-workload-creds.md
+++ b/draft-ietf-wimse-workload-creds.md
@@ -85,7 +85,7 @@ In this document, two credentials are defined:
 
 The Workload Identity Token is targeted for application-layer protocols. The Workload Identity Certificate is targeted for transport-layer protocols. This does not preclude the use of the WIT in transport-layer protocols or the WIC in application-layer protocols, but these are the primary intended uses.
 
-The various protocol bindings that use these credentials to authenticate workloads to each other are specified in separate documents and are out of scope for this document. At the time of writing, three such protocol bindings are defined:
+The various protocol bindings that use these credentials to authenticate workloads to each other are specified in separate documents and are out of scope for this document. Each such binding defines how a workload proves possession of the key material bound to its credential in order to authenticate to another workload. At the time of writing, three such protocol bindings are defined:
 
 * Transport-layer authentication using mutual TLS with the Workload Identity Certificate, specified in {{?I-D.ietf-wimse-mutual-tls}}.
 
@@ -141,7 +141,7 @@ The high-level message flow is as follows:
 2. A transport connection is set up. This may already include the use of the Workload Identity Certificate with transport-layer security, such as TLS.
 3. Workload A prepares to call Workload B. This may include adding application-layer authentication information, such as the Workload Identity Token and proof of possession. Workload B authenticates Workload A.
 4. Workload B authorizes the call. This policy enforcement (Policy Enforcement Point, PEP) may include consulting with an external server (Policy Decision Point, PDP) when making this decision.
-5. Workload B returns a response to Workload A, which may be an error response or a regular one.
+5. Workload B returns a response to Workload A, which may be either a success or an error response.
 
 Depending on the protocol, the workload authentication may happen during step (2) at the transport-layer or at step (3) at the application-layer, or both.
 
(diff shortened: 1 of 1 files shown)


bc-pi (approved):

draft-ietf-wimse-workload-creds.md
```diff
@@ -150,22 +150,23 @@ Depending on the protocol, the workload authentication may happen during step (2
 The Workload Identifier is a URI and its baseline syntax and processing requirements are defined in {{!WIMSE-ID=I-D.ietf-wimse-identifier}}.
 While deployments define how they assign identifiers and what the path portion means, implementations MUST enforce the URI requirements outlined in {{Section 4.1 of WIMSE-ID}}.
 
-Prior to WIMSE, many use cases did not allow for fully granular authentication in containerized runtime platforms.
-For instance, with mutual TLS,
-there's often no clear way to map the request's external access reference
-(e.g., Kubernetes Ingress path, service name, or host header)
-to the SubjectAltName value in the server certificate. This means that the client could only verify
-if the server certificate is valid within a trust domain, not if it's tied to a specific workload.
+Prior to WIMSE, many containerized runtime platforms could not authenticate a peer as a specific workload;
```
arndt-s, line 153: I think that's a bit strong. Do you have a runtime platform in mind that doesn't allow authenticating peers?

draft-ietf-wimse-workload-creds.md
```diff
@@ -150,22 +150,23 @@ Depending on the protocol, the workload authentication may happen during step (2
...
+(such as a Kubernetes Ingress path, service name, or HTTP Host header field)
+to the SubjectAltName in the presented certificate.
+As a result, the client can verify that the server certificate is valid within a trust domain,
+but not that it belongs to the particular workload the client intended to reach.
 
 To enable mutual and granular authentication between workloads, two things must be in place:
 
 - Each workload must know its own identifier.
-- There needs to be an explicit mapping from the external handle used to access a workload (such as an Ingress path or service DNS name)
+- There needs to be an explicit mapping from the external access name used to access a workload (such as an Ingress path or service DNS name)
```
arndt-s, line 164: Isn't this and line 169 saying the same thing?

draft-ietf-wimse-workload-creds.md
```diff
@@ -424,11 +428,11 @@ Teleport - Machine & Workload Identity
 
 Workload Identifiers ({{WIMSE-ID}}) are scoped to a trust domain (the URI authority component) and MUST be interpreted in that trust domain context. Using a Workload Identifier without taking into account the trust domain could allow one domain to issue tokens to spoof identities in another domain. See {{trust-anchors}} for binding trust domains to issuers and trust anchors.
 
-When a deployment uses the `iss` claim for key distribution as described in {{wit-iss-note}}, validators MUST enforce an allowlist of accepted issuers. Absent such a restriction, any entity could stand up an issuer, present a WIT with that issuer's `iss` value, and have it accepted by a validator that fetches and trusts the corresponding key material without verifying the issuer's legitimacy.
+When a deployment uses the `iss` claim for key distribution as required in {{wit-iss-note}}, validators MUST enforce an allowlist of accepted issuers. Absent such a restriction, any entity could stand up an issuer, present a WIT with that issuer's `iss` value, and have it accepted by a validator that fetches and trusts the corresponding key material without verifying the issuer's legitimacy.
```
arndt-s: There's no requirement in wit-iss-note about it. If anything it says "MAY".

draft-ietf-wimse-workload-creds.md
yaronf: The section title is granularity, and I think this sentence (propose by Pieter) is formally correct. It does not say "could not authenticate a peer", it says "could not authenticate a peer *as a specific workload*". And the following text presents an example.

draft-ietf-wimse-workload-creds.md
yaronf: This whole section will be redone in PR #246, so I wouldn't touch it now.

draft-ietf-wimse-workload-creds.md
yaronf: The first line currently says "It is RECOMMENDED that the WIT carries an iss claim", so a SHOULD.

arndt-s: In the context of key discovery it's a MAY, isn't it? The RECOMMENDED is about the iss claim itself.

yaronf: You are right, and the text in iss-note is really hard to parse. Even a section break would improve it.

yaronf:
```suggestion
When a deployment uses the `iss` claim for key distribution as described in {{wit-iss-note}}, validators MUST enforce an allowlist of accepted issuers. Absent such a restriction, any entity could stand up an issuer, present a WIT with that issuer's `iss` value, and have it accepted by a validator that fetches and trusts the corresponding key material without verifying the issuer's legitimacy.
```

Undo the change.

draft-ietf-wimse-workload-creds.md
yaronf:
```suggestion
It is RECOMMENDED that the WIT carries an `iss` claim for the auditing and operational uses described above. Validators are not required to use `iss` when validating the WIT or establishing the workload identity: the trust domain and workload identity are carried in the mandatory `sub` claim ({{WIMSE-ID}}).

Implementations MAY include the `iss` claim in the form of a `https` URL to facilitate key distribution via mechanisms like the `jwks_uri` from {{!RFC8414}}, only when the issuer and verification keys (or metadata such as `jwks_uri`) are configured out of band as in {{trust-anchors}}. Alternative key distribution methods may use only the trust domain from the `sub` claim.
```

Clarifying the section by breaking it into two paragraphs.

draft-ietf-wimse-workload-creds.md
arndt-s: > Prior to WIMSE, many containerized runtime platforms could not authenticate a peer as a specific workload; 

If a peer is within the platform it definitely can. If it's coming from the outside it doesn't.

> they could only establish that the peer belonged to a given trust domain.

I don't really understand that part. 

> With mutual TLS (mTLS), for example, there is often no reliable way to map the external access name a client uses to reach a workload (such as a Kubernetes Ingress path, service name, or HTTP Host header field) to the SubjectAltName in the presented certificate. As a result, the client can verify that the server certificate is valid within a trust domain, but not that it belongs to the particular workload the client intended to reach.

I'm reading this one over and over and I'm not sure I understand it. Is it saying that the client cannot map from requestURI to SubjectAltName? But does it need to? And if it cannot map, how can it tell that's part of a trust domain? And after all, are we actually fixing this?

yaronf: To address only your last question, my understanding of the latest version of the mtls draft is that we're adding an *optional* workload ID to the cert, so you can get granular authz within closed WIMSE environments. It's optional, to accommodate  cross-domain and maybe older environments.

bc-pi: FWIW I tend to agree with Arndt here. 

The first sentence "Prior to WIMSE, many containerized runtime platforms could not authenticate a peer as a specific workload" did seem rather strong but "many" was enough of a qualifier to make me unsure and not push back on it.

I don't think I understand the other part either but it seemed to be text that was mostly their already so didn't seem like a good place to question it.

PR #278, opened 2026-07-27
Add wimse-sign-response for mandated signed responses

### LINKED ISSUE #277: Allow requestor to mandate a signed response
Following a comment from @jhoyla, specifying that response signing (and enforcing/validating this signature) is subject to local policy is brittle. One way to strengthen this behavior is for the requestor to signal that response signing is required. Of course the signal itself must be protected.

In more detail: add a Boolean on the request’s Signature-Input, covered by the signature: `wimse-sign-response=true`.

Pros: integrity-protected; rides with the PoP; hard for proxies to strip without breaking the signature
Cons: new IANA signature-metadata param; server must parse the incoming `Signature-Input` before deciding to sign

yaronf:
## Summary
* Add Boolean request signature parameter `wimse-sign-response` so clients can require a signed response (integrity-protected via Signature-Input).
* Define when response signing is required (parameter or local policy), what to do if the server cannot sign, and regenerate examples accordingly.

Fixes #277


draft-ietf-wimse-http-signature.md
```diff
@@ -338,19 +365,20 @@ receiver, and any TLS-terminating middleboxes that process the traffic.
 ## Authentication
 
 * A workload receiving a request can validate that it is signed correctly, and can identify the sender.
-* A workload receiving a response can similarly authenticate its sender, provided that optional response signing has
-been activated and likewise, the recipient validates this signature.
+* A workload receiving a response can similarly authenticate its sender when response signing is required
+({{signing-the-response}}) and the recipient validates this signature.
```
jsalowey: The and validates this signature seems like odd phrasing or a bit redundant.

```suggestion
* A workload receiving a response can similarly validate the signature and identify the sender when response signing is required
({{signing-the-response}}).
```

jsalowey (approved):

PR #285, opened 2026-08-10
clarify multiple identifiers

### LINKED ISSUE #255: Section 4 - Presence of additional identifiers in WITs

arndt-s:
Resolves #255 by making it more clear that other identifiers are allowed in WIC and WIT.

### DIFF
--- a/draft-ietf-wimse-workload-creds.md
+++ b/draft-ietf-wimse-workload-creds.md
@@ -189,9 +189,9 @@ For Workload Identity Certificates, validators use the trust anchors for the pee
 
 Each WIMSE credential carries exactly one Workload Identifier ({{WIMSE-ID}}) that is used for authentication and authorization. Requiring a single identifier per credential keeps authorization and auditing unambiguous and aligns with common practice for workload X.509 certificates (for example, a single URI in an X.509-SVID).
 
-For a Workload Identity Token, that identifier is the value of the `sub` claim. For a Workload Identity Certificate, that identifier is carried in the single URI SubjectAltName described in {{to-wic}}.
+For a Workload Identity Token, that identifier is the value of the `sub` claim. For a Workload Identity Certificate, that identifier is carried in the single URI SubjectAltName described in {{to-wic}}. Deployments that distinguish several labels for the same runtime (for example a stable service identity and a specific instance) MUST place only one identity into it.
 
-Deployments that distinguish several labels for the same runtime (for example a stable service identity and a specific instance) MUST place the identity required for the access decision in that one credential. Additional correlation for logging or operations MUST NOT be encoded as a second Workload Identifier in the same WIT or WIC. This restriction applies only to Workload Identifiers; deployments MAY use other mechanisms for correlation, interoperability, or integration purposes, such as the optional `jti` claim, a DNSName SubjectAltName, separate credentials, or additional claims as described in {{add-claims}}.
+Additional correlation for logging or operations MUST NOT be encoded as a second Workload Identifier in the same WIT or WIC. This restriction applies only to Workload Identifiers; deployments MAY use other mechanisms for correlation, interoperability, or integration purposes, such as the optional `jti` claim, a DNSName SubjectAltName, separate credentials, or additional claims as described in {{add-claims}}.
 
 # Application-Layer Workload-to-Workload Authentication {#app-layer}
 
--- a/draft-ietf-wimse-workload-creds.md
+++ b/draft-ietf-wimse-workload-creds.md
@@ -191,7 +191,7 @@ Each WIMSE credential carries exactly one Workload Identifier ({{WIMSE-ID}}) tha
 
 For a Workload Identity Token, that identifier is the value of the `sub` claim. For a Workload Identity Certificate, that identifier is carried in the single URI SubjectAltName described in {{to-wic}}.
 
-Deployments that distinguish several labels for the same runtime (for example a stable service identity and a specific instance) MUST place the identity required for the access decision in that one credential. Additional correlation for logging or operations MUST NOT be encoded as a second Workload Identifier in the same WIT or WIC. Deployments MAY use other mechanisms, such as the optional `jti` claim, separate credentials, or additional claims as described in {{add-claims}}.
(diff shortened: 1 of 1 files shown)


draft-ietf-wimse-workload-creds.md
```diff
@@ -191,7 +191,7 @@ Each WIMSE credential carries exactly one Workload Identifier ({{WIMSE-ID}}) tha
...
 
-Deployments that distinguish several labels for the same runtime (for example a stable service identity and a specific instance) MUST place the identity required for the access decision in that one credential. Additional correlation for logging or operations MUST NOT be encoded as a second Workload Identifier in the same WIT or WIC. Deployments MAY use other mechanisms, such as the optional `jti` claim, separate credentials, or additional claims as described in {{add-claims}}.
+Deployments that distinguish several labels for the same runtime (for example a stable service identity and a specific instance) MUST place the identity required for the access decision in that one credential. Additional correlation for logging or operations MUST NOT be encoded as a second Workload Identifier in the same WIT or WIC. This restriction applies only to Workload Identifiers; a credential MAY carry other identifiers, such as the `jti` claim or a DNSName SubjectAltName, for correlation, interoperability, or integration purposes. Deployments MAY use other mechanisms, such as the optional `jti` claim, separate credentials, or additional claims as described in {{add-claims}}.
```
yaronf: This is confusing: `jti` appears in both the 3rd and the last sentence. I think the two sentences should be consolidated.

arndt-s: Updated, please take a look

PieterKas: Is this really about the value that is acceptable as a sub claim in the WIT? Perhaps something like this (although this does not quite work for WICs).

```suggestion
Deployments that distinguish several labels for the same runtime (for example a stable service identity and a specific instance) MUST use the identity required for the access decision as the `sub` claim in the credential. Additional correlation for logging or operations MUST NOT be encoded as a second Workload Identifier in the `sub` claim in the same WIT or WIC. This restriction applies only to Workload Identifiers; a credential MAY carry other identifiers, such as the `jti` claim or a DNSName SubjectAltName, for correlation, interoperability, or integration purposes. Deployments MAY use other mechanisms, such as the optional `jti` claim, separate credentials, or additional claims as described in {{add-claims}}.
```

arndt-s: I believe that context is already given by the previous paragraph.

> For a Workload Identity Token, that identifier is the value of the `sub` claim. For a Workload Identity Certificate, that identifier is carried in the single URI SubjectAltName described in {{to-wic}}.

yaronf (approved):

bc-pi (commented):
wfm, thanks @arndt-s

bc-pi (approved):

draft-ietf-wimse-workload-creds.md
PieterKas: It did not carry over for me, so perhaps just being explicit helps other readers (reading"that credential" did not build the bridge to the previous paragraph - but I was also in a rush reading it)? Alternatively reference back to the previous paragraph perhaps?

The term Worklod Identitfier itself becomes nuanced - I can imagine someone using a WIMSE/SPIFFE workload identifier for the 'sub' claim and then also using it for another identifier (or maybe just using a URI). so being explicit about which claim carries the workload identifier is perhaps the real thing to make sure readers understand.

arndt-s: @PieterKas I moved the sentence in question into the prior paragraph. I think that makes it a bit more clear. WDYT?

draft-ietf-wimse-workload-creds.md
```diff
@@ -189,9 +189,9 @@ For Workload Identity Certificates, validators use the trust anchors for the pee
 
 Each WIMSE credential carries exactly one Workload Identifier ({{WIMSE-ID}}) that is used for authentication and authorization. Requiring a single identifier per credential keeps authorization and auditing unambiguous and aligns with common practice for workload X.509 certificates (for example, a single URI in an X.509-SVID).
 
-For a Workload Identity Token, that identifier is the value of the `sub` claim. For a Workload Identity Certificate, that identifier is carried in the single URI SubjectAltName described in {{to-wic}}.
+For a Workload Identity Token, that identifier is the value of the `sub` claim. For a Workload Identity Certificate, that identifier is carried in the single URI SubjectAltName described in {{to-wic}}. Deployments that distinguish several labels for the same runtime (for example a stable service identity and a specific instance) MUST place only one identity into it.
```
PieterKas: suggest replacing "identity" with "identifier'

```suggestion
For a Workload Identity Token, that identifier is the value of the `sub` claim. For a Workload Identity Certificate, that identifier is carried in the single URI SubjectAltName described in {{to-wic}}. Deployments that distinguish several labels for the same runtime (for example a stable service identity and a specific instance) MUST place only one identifier into it.
```

draft-ietf-wimse-workload-creds.md
PieterKas: The change works, I made on editorial suggestion.

Issue #272, opened 2026-07-16
followup on Editorial fixes for workload-creds #267


bc-pi:
FWIW I tend to agree with Arndt here. 

The first sentence "Prior to WIMSE, many containerized runtime platforms could not authenticate a peer as a specific workload" did seem rather strong but "many" was enough of a qualifier to make me unsure and not push back on it.

I don't think I understand the other part either but it seemed to be text that was mostly their already so didn't seem like a good place to question it.

_Originally posted by @bc-pi in https://github.com/ietf-wg-wimse/draft-ietf-wimse-s2s-protocol/pull/267#discussion_r3595285661_

PR #286, opened 2026-08-10
Clarify workload identifier and add security considerations

arndt-s:
Added security considerations for granular auth section.

Related to #272

### DIFF
--- a/draft-ietf-wimse-workload-creds.md
+++ b/draft-ietf-wimse-workload-creds.md
@@ -436,9 +436,7 @@ When a deployment uses the `iss` claim for key distribution as described in {{wi
 
 A workload that validates its peer's credential against the trust anchors for a trust domain and then stops has learned only that its peer belongs to that trust domain. It has not learned which workload the peer is. {{granular-auth}} requires the comparison that closes this gap.
 
-Without that comparison, every workload in the trust domain looks the same. Any workload holding a valid credential can be accepted in place of any other, so an attacker who compromises one workload gains the access of all of them, even if the workload they compromised had little access of its own. A credential issued by mistake, or one belonging to a workload that no longer exists, can be used in the same way until it expires.
-
-Both sides of a call are affected. A client that does not check which workload it reached can be tricked into sending its requests, along with its own credential and proof of possession, to a workload the attacker controls in the same trust domain. A receiving application that only checks trust domain membership gives every workload in the domain the access it meant to give to one.
+Without that comparison, every workload in the trust domain looks the same. Any workload holding a valid credential can be accepted in place of any other, so an attacker who compromises one workload gains the access of all of them, even if the workload they compromised had little access of its own. A credential issued by mistake, or one belonging to a workload that no longer exists, can be used in the same way until it expires. This affects both sides of a call, workloads acting a clients and workloads acting as servers.
 
 Proof of possession does not help here. It shows that the peer holds the private key for the credential it presented ({{wit-pop}}). It does not show that this credential is the one the peer was supposed to present. The same is true when a Workload Identity Certificate is used at the transport layer.
 
--- a/draft-ietf-wimse-workload-creds.md
+++ b/draft-ietf-wimse-workload-creds.md
@@ -150,18 +150,24 @@ Depending on the protocol, the workload authentication may happen during step (2
 The Workload Identifier is a URI and its baseline syntax and processing requirements are defined in {{!WIMSE-ID=I-D.ietf-wimse-identifier}}.
 While deployments define how they assign identifiers and what the path portion means, implementations MUST enforce the URI requirements outlined in {{Section 4.1 of WIMSE-ID}}.
 
-A Workload Identifier is generally not a name that a client dials. Clients reach a workload through an external access name,
-such as a DNS name or a Kubernetes Service or Ingress host, while the authority component of a Workload Identifier is a
-trust domain rather than a resolvable host. Validating a peer's credential against the trust anchors of its trust domain
-({{trust-anchors}}) therefore establishes only that the peer holds a credential issued within that domain, not that it is
-the workload the client intended to reach.
+Prior to WIMSE, many containerized runtime platforms could not authenticate a peer as a specific workload;  
+they could only establish that the peer belonged to a given trust domain.
+With mutual TLS (mTLS), for example, there is often no reliable way to map the external access name a client uses to reach a workload
+(such as a Kubernetes Ingress path, service name, or HTTP Host header field)
+to the SubjectAltName in the presented certificate.
+As a result, the client can verify that the server certificate is valid within a trust domain,
+but not that it belongs to the particular workload the client intended to reach.
 
-To authenticate a specific peer, a client MUST compare the Workload Identifier in the presented credential against the
-identifier it expects for the workload it is calling; successful trust anchor validation alone MUST NOT be treated as
-authentication of that workload (see {{trust-domain-membership}} for security considerations). This requires a mapping from
-the external access name to the expected Workload Identifier, and agreement between the peers on the granularity of
-identifiers, for example one per service or one per running instance. This document does not define how that mapping is
-established or distributed.
+To enable mutual and granular authentication between workloads, two things must be in place:
+
+- Each workload must know its own identifier.
+- There needs to be an explicit mapping from the external access name used to access a workload (such as an Ingress path or service DNS name)
+to its Workload Identifier.
+
+Once these conditions are met, the methods described in this document can be used for the caller and callee to mutually authenticate.
+
+Implementations MUST allow for defining this mapping between the workload's external access name and the Workload Identifier (e.g., through
+callback functions). Deployments SHOULD use these features to establish a consistent set of identifiers within their environment.
 
 # Conventions and Definitions
 
@@ -566,7 +572,7 @@ IANA is requested to register the following entries to the "Hypertext Transfer P
 
 ## draft-ietf-wimse-workload-creds-03
 
-* Rework the Workload Identifiers and Authentication Granularity section and added a new Security Considerations subsection to describe the impact.
+* Added a section to describe the security considerations of granular authentication.
 * Editorial: consistent use of "proof of possession"/"PoP", with the abbreviation expanded on first use, and consistent capitalization of the defined term "Workload Identifier".
 
 ## draft-ietf-wimse-workload-creds-02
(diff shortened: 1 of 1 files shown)


draft-ietf-wimse-workload-creds.md
```diff
@@ -150,24 +150,18 @@ Depending on the protocol, the workload authentication may happen during step (2
 The Workload Identifier is a URI and its baseline syntax and processing requirements are defined in {{!WIMSE-ID=I-D.ietf-wimse-identifier}}.
 While deployments define how they assign identifiers and what the path portion means, implementations MUST enforce the URI requirements outlined in {{Section 4.1 of WIMSE-ID}}.
 
-Prior to WIMSE, many containerized runtime platforms could not authenticate a peer as a specific workload;
```
yaronf, line 153: I think this context/motivation is important.

draft-ietf-wimse-workload-creds.md
```diff
@@ -150,24 +150,18 @@ Depending on the protocol, the workload authentication may happen during step (2
...
-
-Implementations MUST allow for defining this mapping between the workload's external access name and the Workload Identifier (e.g., through
-callback functions). Deployments SHOULD use these features to establish a consistent set of identifiers within their environment.
+A Workload Identifier is generally not a name that a client dials. Clients reach a workload through an external access name,
+such as a DNS name or a Kubernetes Service or Ingress host, while the authority component of a Workload Identifier is a
+trust domain rather than a resolvable host. Validating a peer's credential against the trust anchors of its trust domain
+({{trust-anchors}}) therefore establishes only that the peer holds a credential issued within that domain, not that it is
+the workload the client intended to reach.
+
+To authenticate a specific peer, a client MUST compare the Workload Identifier in the presented credential against the
```
yaronf: In the case of WPT, most of the authentication happens on the *server* side, by (1) validating the incoming WIT/WPT and (2) verifying the audience.

yaronf (changes requested):

yaronf:
This is an alternative text to the still open #246 - we should discuss both of them together.

draft-ietf-wimse-workload-creds.md
```diff
@@ -150,24 +150,18 @@ Depending on the protocol, the workload authentication may happen during step (2
...
+identifier it expects for the workload it is calling; successful trust anchor validation alone MUST NOT be treated as
+authentication of that workload (see {{trust-domain-membership}} for security considerations). This requires a mapping from
```
yaronf: My proposed text avoids the "mapping" notion (which is hard to define precisely) and instead focuses on the information that needs to be available to the endpoints so that each of them can enforce its security requirements.

draft-ietf-wimse-workload-creds.md
arndt-s: I'm not sure I agree with that statement though. It's also that we don't really solve the problem. The mapping between location identifier and workload identifier is out of scope.

draft-ietf-wimse-workload-creds.md
```diff
@@ -150,24 +150,18 @@ Depending on the protocol, the workload authentication may happen during step (2
...
-
-- Each workload must know its own identifier.
-- There needs to be an explicit mapping from the external access name used to access a workload (such as an Ingress path or service DNS name)
-to its Workload Identifier.
-
-Once these conditions are met, the methods described in this document can be used for the caller and callee to mutually authenticate.
-
-Implementations MUST allow for defining this mapping between the workload's external access name and the Workload Identifier (e.g., through
-callback functions). Deployments SHOULD use these features to establish a consistent set of identifiers within their environment.
+A Workload Identifier is generally not a name that a client dials. Clients reach a workload through an external access name,
```
bc-pi: I realize the wording is difficult (see the terrible words in my suggestion below) but I think "dials" might be a little too colloquial for a document that aspires to be an RFC. 

```suggestion
A Workload Identifier is generally not a name that a client uses to initiate invocation of a service. Clients reach a workload through an external access name,
```

draft-ietf-wimse-workload-creds.md
bc-pi: Similar to https://github.com/ietf-wg-wimse/draft-ietf-wimse-s2s-protocol/pull/267#discussion_r3595285661 and #272 which it begat, I tend to agree with @arndt-s here.

arndt-s:
@bc-pi @yaronf as discussed on Tuesday, I removed the main part and kept the security considerations.

draft-ietf-wimse-workload-creds.md
```diff
@@ -150,7 +150,7 @@ Depending on the protocol, the workload authentication may happen during step (2
 The Workload Identifier is a URI and its baseline syntax and processing requirements are defined in {{!WIMSE-ID=I-D.ietf-wimse-identifier}}.
 While deployments define how they assign identifiers and what the path portion means, implementations MUST enforce the URI requirements outlined in {{Section 4.1 of WIMSE-ID}}.
 
-Prior to WIMSE, many containerized runtime platforms could not authenticate a peer as a specific workload;
+Prior to WIMSE, many containerized runtime platforms could not authenticate a peer as a specific workload;
```
bc-pi: I am not well versed in containerized runtime platforms but this sentence and the paragraph below were the impetus for #272.

draft-ietf-wimse-workload-creds.md
```diff
@@ -432,6 +432,16 @@ Workload Identifiers ({{WIMSE-ID}}) are scoped to a trust domain (the URI author
...
 
+## Trust Domain Membership and Peer Authentication {#trust-domain-membership}
+
+A workload that validates its peer's credential against the trust anchors for a trust domain and then stops has learned only that its peer belongs to that trust domain. It has not learned which workload the peer is. {{granular-auth}} requires the comparison that closes this gap.
+
+Without that comparison, every workload in the trust domain looks the same. Any workload holding a valid credential can be accepted in place of any other, so an attacker who compromises one workload gains the access of all of them, even if the workload they compromised had little access of its own. A credential issued by mistake, or one belonging to a workload that no longer exists, can be used in the same way until it expires.
+
+Both sides of a call are affected. A client that does not check which workload it reached can be tricked into sending its requests, along with its own credential and proof of possession, to a workload the attacker controls in the same trust domain. A receiving application that only checks trust domain membership gives every workload in the domain the access it meant to give to one.
```
yaronf, line 441: I would add a comment at the end of this paragraph that the exact security properties (who authenticates whom etc.) are dependent on the WIMSE mechanism being used (MTLS vs. http-sig vs. WPT).

draft-ietf-wimse-workload-creds.md
```diff
@@ -432,6 +432,16 @@ Workload Identifiers ({{WIMSE-ID}}) are scoped to a trust domain (the URI author
...
+
+Proof of possession does not help here. It shows that the peer holds the private key for the credential it presented ({{wit-pop}}). It does not show that this credential is the one the peer was supposed to present. The same is true when a Workload Identity Certificate is used at the transport layer.
```
yaronf, line 443: I disagree with this paragraph. Take the (reasonable) assumptions that the credential was correctly issued (that the workload possessed the private key at that time), *and* that private keys are hard to steal. Then the PoP does show that the credential is correct.

yaronf (changes requested):

draft-ietf-wimse-workload-creds.md
arndt-s: This PR doesn't target #272 anymore, see #246

bc-pi: I have seen #246 (more than I care for) and it also doesn't address #272.

draft-ietf-wimse-workload-creds.md
arndt-s: This section is about situations where a peer only validates a credential against the trust domain anchor and basically ignores the identifier for access control. Every workload will be the same.

draft-ietf-wimse-workload-creds.md
arndt-s: I don't think proof of possession matters here. It's generic, no matter what PoP mechanism is used.

```suggestion
Both sides of a call are affected. A receiving application that only checks trust domain membership gives every workload in the domain the access it meant to give to one. The same applies when a server presents a Workload Identity Certificate and a client only validates the trust domain membership. It does not apply when DNS SAN is used.
```

arndt-s: Discussed in the call today and removed that section. Instead, the prior section received a "both sides of a call are affected" sentence.

draft-ietf-wimse-workload-creds.md
```diff
@@ -440,6 +440,14 @@ Workload Identifiers ({{WIMSE-ID}}) are scoped to a trust domain (the URI author
 
 When a deployment uses the `iss` claim for key distribution as described in {{wit-iss-note}}, validators MUST enforce an allowlist of accepted issuers. Absent such a restriction, any entity could stand up an issuer, present a WIT with that issuer's `iss` value, and have it accepted by a validator that fetches and trusts the corresponding key material without verifying the issuer's legitimacy.
 
+## Trust Domain Membership and Peer Authentication {#trust-domain-membership}
+
+A workload that validates its peer's credential against the trust anchors for a trust domain and then stops has learned only that its peer belongs to that trust domain. It has not learned which workload the peer is. {{granular-auth}} requires the comparison that closes this gap.
```
yaroslavros:
```suggestion
Validating a peer's credential against the trust anchors for a trust domain establishes that the peer belongs to that trust domain, but does not by itself establish which workload the peer represents. The additional identity comparison described in {{granular-auth}} is required to make that distinction.
```

draft-ietf-wimse-workload-creds.md
```diff
@@ -440,6 +440,14 @@ Workload Identifiers ({{WIMSE-ID}}) are scoped to a trust domain (the URI author
...
+
+Without that comparison, every workload in the trust domain looks the same. Any workload holding a valid credential can be accepted in place of any other, so an attacker who compromises one workload gains the access of all of them, even if the workload they compromised had little access of its own. A credential issued by mistake, or one belonging to a workload that no longer exists, can be used in the same way until it expires. This affects both sides of a call, workloads acting as clients and workloads acting as servers.
```
yaroslavros:
```suggestion
Without that comparison workloads in the trust domain cannot be securely distinguished. Any workload holding a valid credential can be accepted in place of any other, so an attacker compromising one workload gains can move laterally and gain access on behalf of any other workload within given trust domain. A credential issued by mistake, or one belonging to a workload that no longer exists, can be used in the same way until it expires. This affects both sides of a call, workloads acting as clients and workloads acting as servers.
```

draft-ietf-wimse-workload-creds.md
```diff
@@ -440,6 +440,14 @@ Workload Identifiers ({{WIMSE-ID}}) are scoped to a trust domain (the URI author
...
+
+Proof of possession does not help here. It shows that the peer holds the private key for the credential it presented ({{wit-pop}}). It does not show that this credential is the one the peer was supposed to present. The same is true when a Workload Identity Certificate is used at the transport layer.
```
yaroslavros:
```suggestion
Proof of possession does not mitigate this vulnerability. It shows that the peer holds the private key for the credential it presented ({{wit-pop}}), but it does not show that this credential is the one the peer was supposed to present. The same consideration applies when a Workload Identity Certificate is used at the transport layer.
```

yaroslavros (approved):

PR #291, opened 2026-08-19
Add WIT validation procedure for recipients

### LINKED ISSUE #290: Guidance for validating the WIT
The creds draft seems to me missing guidance for validating the WIT.  Both the WIT and HTTP signatures drafts will need to reference it.

yaronf:
## Summary
* Add Section 5.1.4 in workload-creds with a MUST checklist for validating a presented WIT (signature, `typ`, `sub`, `exp`, `cnf`).
* Point WPT and HTTP signatures at that procedure instead of restating WIT checks.

Fixes #290

### DIFF
--- a/draft-ietf-wimse-workload-creds.md
+++ b/draft-ietf-wimse-workload-creds.md
@@ -181,7 +181,7 @@ This document expects that trust anchors used to validate Workload Identity Cert
 
 This document does not define a mechanism for discovering trust anchor information. Consumers MUST bind each trust domain to an authorized issuer (or set of issuers) and to the corresponding trust anchors. That mapping MUST be distributed via a secure out-of-band mechanism. In particular, for a WIT the `iss` claim (the JWT issuer claim {{RFC7519}}, defined in {{to-wit}}) MUST NOT be used to look up trust anchor material directly from information carried only in the token (such as by fetching a JWKS from a URL found only in the WIT).
 
-To validate a WIT, the receiving workload verifies the JWS signature using key material from the trust anchors configured for the trust domain in the `sub` claim. Trust anchor material for WIT validation MAY include more than one public key so that operators can rotate signing keys. Verification follows {{RFC7515}} and the JWT best practices in {{RFC8725}}. In practice, the `alg` header parameter would identify an algorithm that the validator accepts for that trust domain, and the signature would be checked under a key drawn from the configured trust anchors. If the JWS header contains a `kid` parameter, the validator would use the key in the trust anchor material with the same `kid` value. If `kid` is absent, deployment policy would need to provide unambiguous key selection (for example, when only a single key is configured for that trust domain). Recipient processing of a presented WIT is specified in {{wit-validation}}.
+To validate a WIT, the receiving workload verifies the JWS signature using key material from the trust anchors configured for the trust domain in the `sub` claim. Trust anchor material for WIT validation MAY include more than one public key so that operators can rotate signing keys. In practice, the `alg` header parameter would identify an algorithm that the validator accepts for that trust domain, and the signature would be checked under a key drawn from the configured trust anchors. If the JWS header contains a `kid` parameter, the validator would use the key in the trust anchor material with the same `kid` value. If `kid` is absent, deployment policy would need to provide unambiguous key selection (for example, when only a single key is configured for that trust domain). Recipient processing of a presented WIT is specified in {{wit-validation}}.
 
 For Workload Identity Certificates, validators use the trust anchors for the peer's trust domain to validate the X.509 chain and the Workload Identifier in the certificate, as described in {{to-wic}} and in companion specifications such as {{!I-D.ietf-wimse-mutual-tls}}.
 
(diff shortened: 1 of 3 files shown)


draft-ietf-wimse-workload-creds.md
```diff
@@ -340,6 +340,19 @@ It is RECOMMENDED that the WIT carries an `iss` claim for the auditing and opera
 
 Implementations MAY include the `iss` claim in the form of a `https` URL to facilitate key distribution via mechanisms like the `jwks_uri` from {{!RFC8414}}, only when the issuer and verification keys (or metadata such as `jwks_uri`) are configured out of band as in {{trust-anchors}}. Alternative key distribution methods may use only the trust domain from the `sub` claim.
 
+### Validating the WIT {#wit-validation}
```
arndt-s, line 343: This now stands in contrast to what is said in the section above:

> Verification follows {{RFC7515}} and the JWT best practices in {{RFC8725}}.

draft-ietf-wimse-workload-creds.md
```diff
@@ -340,6 +340,19 @@ It is RECOMMENDED that the WIT carries an `iss` claim for the auditing and opera
...
+
+To validate a WIT, the recipient MUST ensure the following:
+
+* The token is a single well-formed JWT using JWS compact serialization ({{RFC7515}}, {{RFC7519}}).
+* The `typ` JOSE header parameter conveys a media type of `wit+jwt`.
+* The `alg` JOSE header parameter identifies an asymmetric digital signature algorithm that is acceptable per the trust domain's policy; the value `none` MUST NOT be accepted. Algorithms used with symmetric keys and encryption algorithms MUST NOT be accepted.
+* The signature is valid using key material from the trust anchors configured for the trust domain in the `sub` claim, as described in {{trust-anchors}}. If the JWS header contains a `kid` parameter, the recipient MUST use the configured key with that `kid` value. If `kid` is absent, key selection MUST follow the deployment policy in {{trust-anchors}}.
+* The `sub` claim is present and is a single Workload Identifier ({{WIMSE-ID}}, {{single-workload-identity}}) whose trust domain matches the trust anchors used.
```
arndt-s, line 351: Prior step already covers this. Maybe we need to move this one step up?

draft-ietf-wimse-workload-creds.md
```diff
@@ -340,6 +340,19 @@ It is RECOMMENDED that the WIT carries an `iss` claim for the auditing and opera
 
 Implementations MAY include the `iss` claim in the form of a `https` URL to facilitate key distribution via mechanisms like the `jwks_uri` from {{!RFC8414}}, only when the issuer and verification keys (or metadata such as `jwks_uri`) are configured out of band as in {{trust-anchors}}. Alternative key distribution methods may use only the trust domain from the `sub` claim.
 
+### Validating the WIT {#wit-validation}
+
+To validate a WIT, the recipient MUST ensure the following:
+
+* The token is a single well-formed JWT using JWS compact serialization ({{RFC7515}}, {{RFC7519}}).
+* The `typ` JOSE header parameter conveys a media type of `wit+jwt`.
+* The `alg` JOSE header parameter identifies an asymmetric digital signature algorithm that is acceptable per the trust domain's policy; the value `none` MUST NOT be accepted. Algorithms used with symmetric keys and encryption algorithms MUST NOT be accepted.
```
arndt-s, line 349: This step already requires knowing the trust domain, which is only identified at later steps.

draft-ietf-wimse-workload-creds.md
arndt-s: I now recall a discussion about this. We considered between referencing other drafts for JWT validation or specifying it on our own and potentially duplicating most of it.

I'm not sure if it makes sense to bring all the details about `exp` validation, clock skew and what not into this draft. It also means we need to talk about the security considerations of it. So I'm in favor of just referencing other drafts that focus on that part. 

But I also agree that we're currently a bit too shallow on it. There's specific stuff in here such as the `wit+jwt` type, the trust anchors (somewhat covered in the {{trust-anchors}} section and that the `cnf` claim must be present.

draft-ietf-wimse-workload-creds.md
```diff
@@ -181,7 +181,7 @@ This document expects that trust anchors used to validate Workload Identity Cert
...
+To validate a WIT, the receiving workload verifies the JWS signature using key material from the trust anchors configured for the trust domain in the `sub` claim. Trust anchor material for WIT validation MAY include more than one public key so that operators can rotate signing keys. Verification follows {{RFC7515}} and the JWT best practices in {{RFC8725}}. In practice, the `alg` header parameter would identify an algorithm that the validator accepts for that trust domain, and the signature would be checked under a key drawn from the configured trust anchors. If the JWS header contains a `kid` parameter, the validator would use the key in the trust anchor material with the same `kid` value. If `kid` is absent, deployment policy would need to provide unambiguous key selection (for example, when only a single key is configured for that trust domain). Recipient processing of a presented WIT is specified in {{wit-validation}}.
```
```suggestion
To validate a WIT, the receiving workload verifies the JWS signature using key material from the trust anchors configured for the trust domain in the `sub` claim. Trust anchor material for WIT validation MAY include more than one public key so that operators can rotate signing keys. Verification follows {{RFC7515}} and the JWT best practices in {{RFC8725}}. In practice, the `alg` header parameter would identify an algorithm that the validator accepts for that trust domain, and the signature would be checked under a key drawn from the configured trust anchors. If the JWS header contains a `kid` parameter, the validator would use the key in the trust anchor material with the same `kid` value. If `kid` is absent, deployment policy would need to provide unambiguous key selection (for example, when only a single key is configured for that trust domain).

Extra validation steps of the WIT in addition to RFC7515 and RFC8725 are:

* The `typ` JOSE header parameter conveys a media type of `wit+jwt`.
* The `sub` is a Workload Identifier and identifies a trust domain.
* The `cnf` claim is present and contains a `jwk` object with an `alg` member as specified in {{to-wit}}.
* The signature is valid based on the rules described above.
```

The last bullet point needs refinement.

draft-ietf-wimse-workload-creds.md
yaronf: Agree. Should be fixed now.

bc-pi (approved):

PR #293, opened 2026-08-25
Try to be more clear on how the various token hash claims are done

### LINKED ISSUE #292: Clarify wth calculation
The `wth` claim is defined as:
```
      -  wth: Hash of the Workload Identity Token, defined in
         [I-D.ietf-wimse-workload-creds].  The value is the base64url
         encoding of the SHA-256 hash of the ASCII encoding of the WIT's
         value.
```

What is the ASCII encoding of the WIT's value?

bc-pi:
related to #292


yaronf (approved):
Approved, though it would be nicer to define base64url once instead of quarce (4 times).

(I so wanted a word like "thrice", and the chat says: there is an archaic/rare “quarce” (also historically fourice), but essentially nobody uses it today.)

sorindumitru (approved):

PR #296, opened 2026-08-27
Add an implementation status entry for wimsey

kanywst:
An entry for `wimsey`, a Rust implementation of these documents, per Yaron's invitation on the list: <https://github.com/kanywst/wimsey>.

The WPT and mutual-TLS documents have no Implementation Status section. `wimsey` implements both — happy to add the RFC 7942 boilerplate and an entry to each if you want it.

`Contact:` is a GitHub link rather than an email, following the SPIFFE entry. Both documents build locally.


yaronf (approved):
Thank you for the PR!

arndt-s (approved):

PR #300, opened 2026-08-28
Remove Workload Identity Key Management section from WPT...

### LINKED ISSUE #298: WPT-02 sec 3
as mentioned in https://mailarchive.ietf.org/arch/msg/wimse/aEhoJl7WUmeREk_bKmrlVxXMk7Q/ the target/context of [3.2. ](https://datatracker.ietf.org/doc/html/draft-ietf-wimse-wpt-02#section-3.2)[Workload Identity Key Management](https://datatracker.ietf.org/doc/html/draft-ietf-wimse-wpt-02#name-workload-identity-key-manag) is confusing to me

bc-pi:
which seems more appropriate in ietf-wimse-workload-creds where it already is 
for #298


arndt-s (approved):
There's probably more of these leftovers from the split

Issue #301, opened 2026-08-29
WGLC: local-policy response signing and signature label selection (Anton Sokolov)


yaronf:
WGLC follow-up on `draft-ietf-wimse-http-signature-06` from @tyche-dev (Anton Sokolov, Tyche Institute). Agrees with the points in #297 and raises two additional issues.

## 1. Local-policy response signing is not client-verifiable

Section 3.4 makes a response signature mandatory both when `wimse-sign-response` is present and when server local policy requires it. In the former case the client can reject an unsigned response. In the latter, the client may not know the policy, so a proxy that strips the signature yields an ordinary unsigned response.

The document should either require such a policy to be advertised or negotiated in authenticated request semantics available to the client, or state that a policy-only requirement does not provide a client-verifiable mandatory-signature guarantee.

Distinct from the voluntary-signing / `wimse-req-nonce` replay issue in #297.

## 2. Do not select the WIMSE signature by label

Section 3's suggestion to use `wimse` as the signature label is incompatible with RFC 9421's treatment of labels (unique identifiers; intermediaries may relabel; applications should not put semantic meaning on labels — see RFC 9421 Sections 4 and 7.2.5). Application selection is what `tag` is for (RFC 9421 Section 7.2.7); this profile already requires `tag="wimse-workload-to-workload"`.

The label guidance should be removed or made explicitly non-semantic. Recipients should select by `tag`, then validate the profile's requirements on the chosen signature (rather than matching on the full covered-component list as a discovery mechanism).

PR #302, opened 2026-08-30
HTTP-Sig WGLC comments from Kieran and Anton

### LINKED ISSUE #297: WGLC: @request-target, voluntary response signing, nonce scope (Kieran Sweeney)
WGLC review of `draft-ietf-wimse-http-signature-06` from @kierandotai (Kieran Sweeney), [mail archive](https://mailarchive.ietf.org/arch/msg/wimse/PhjX6Pc4l7DSqJCBHG5w8fKEg5k/).

Overall: layering and design are sound (`wimse-aud`, forbidding `keyid`/`alg`, credentials out of scope). Not ready yet for the reasons below.

## 1. `@request-target` (blocking)

Section 3 mandates `@request-target` as a covered component. RFC 9421 Section 2.2.5 says this is difficult to construct reliably outside HTTP/1.1 and is NOT RECOMMENDED when other HTTP versions may be in use.

The deployments this draft targets (multi-hop TLS proxies/LBs, often HTTP/2) are exactly where `@request-target` can diverge across hops (origin-form vs absolute-form, version translation), breaking end-to-end verification.

**Suggested fix:** replace `@request-target` / `@request-target;req` with `@path` and `@query` (when a query is present). Recipient binding is already carried by mandatory `wimse-aud` and Section 3.1.

## 2. Voluntary signed responses without `wimse-req-nonce`

Section 3 requires `wimse-req-nonce` only when response signing is required, but Section 3.4 allows the server to sign a response voluntarily. A voluntarily signed response need not carry `wimse-req-nonce`, and the client is not told to check it — a middlebox could replay one client's signed response to another.

**Suggested fix:** require `wimse-req-nonce` on every signed response (request `nonce` is already mandatory).

## 3. Nonce uniqueness scope

Section 3: unique among messages to the same recipient. Section 5.1: unique within the scope of a particular sender. Pick one scope and use it in both places.

yaronf:
5 commits, each addressing one comment.

Closes #297 and #301.


draft-ietf-wimse-http-signature.md
```diff
@@ -139,12 +143,11 @@ The following signature parameters in the `Signature-Input` header MUST NOT be u
 * `alg` - The signature algorithm is specified in the `jwk` section of the `cnf` claim in the WIT. See {{I-D.ietf-wimse-workload-creds}} and Sec. 3.3.7 of {{RFC9421}} for details.
 
 It is RECOMMENDED to include only one signature with the HTTP message.
-If multiple ones are included, then the signature label included in both the `Signature-Input` and `Signature` headers SHOULD
-be `wimse`.
+The WIMSE signature is the one whose `tag` parameter is `wimse-workload-to-workload` ({{Section 7.2.7 of RFC9421}}). If no signature has that `tag` value, the message does not carry a WIMSE HTTP signature. If more than one signature has that `tag` value, the recipient MUST reject the message. When more than one signature is present, recipients MUST use that `tag` to find the WIMSE signature; they MUST NOT choose by label ({{Section 7.2.5 of RFC9421}}).
 
-A sender MUST ensure that each nonce it generates is unique, at least among messages sent to the same recipient.
+A sender MUST ensure that each nonce it generates is unique. Nonces are random and MUST be long enough that the probability of collision among honest senders is negligible. A sender cluster therefore need not share nonce state.
```
jsalowey, line 148: DO nonces need to be random (unpredictable) or just unique?   Also the the bit about sharing nonce state seems ambiguous. 

```suggestion
A sender MUST ensure that each nonce it generates is unique. Nonces MUST be long enough that the probability of collision among honest senders is negligible. A sender cluster therefore needs to make sure servers do not send the same nonce.
```

draft-ietf-wimse-http-signature.md
```diff
@@ -175,32 +178,31 @@ This parameter is not mandatory. Moreover, the server MAY sign the response even
...
+* Local policy at the server requires response signing for that exchange.
 
 If the server is required to sign the response but cannot produce a signed response (for example, because the response is streamed or
 exceptionally large), it MUST NOT return a successful unsigned response; it MUST return an error
 as described in {{error-conditions}}.
 
-If response signing is required, the client (recipient of the response) MUST check that the signature exists and validate it.
-The response MUST be rejected if a signature is absent or fails to validate.
-The client MUST verify that `wimse-req-nonce` matches the `nonce` it included in its request's `Signature-Input`.
+The client MUST reject an unsigned response when it required a signed response, either by sending `wimse-sign-response` or by local configuration. A requirement that exists only at the server cannot be enforced by the client: a stripped signature looks like an ordinary unsigned response.
```
jsalowey, line 201: The local policy seems redundant here.  If the local policy is sent then the client sends wimse-sign-response.  Is there a case where it wouldn't?  
We might what to have a sentence like:
"If the client is configured to required a signed response then it MUST send include the wimse-sign-response parameter with the Boolean value of true. "
then you can leave off the either by sending  wimse-sign-response or by local configuration.

draft-ietf-wimse-http-signature.md
yaronf: Practically it's easiest to mandate that they are random - easier for devs to understand and easier to implement. In fact I would rather add:
> Nonces MUST be generated at random.

The alternative is not only more confusing, it could easily result in complicated solutions where cluster members share state to avoid nonce duplication. This is what my "ambiguous" text tried to address.

jsalowey (approved):

PR #310, opened 2026-09-08
Add an Implementation Status section to the WPT draft

kanywst:
An Implementation Status entry for `wimsey` (<https://github.com/kanywst/wimsey>), which implements the WPT. The document has no Implementation Status section yet, so this adds one, worded as in draft-ietf-wimse-http-signature.

The coverage line states what is not implemented: the `WPT` HTTP authentication scheme introduced in -02, and single-use `jti` tracking, are left to the caller.


bc-pi:
My preference would be not to have an Implementation Status section.

bc-pi (changes requested):
see prior

kanywst:
Understood, dropping this. If a section is added later, I'll contribute the entry then.

PR #311, opened 2026-09-08
Add an Implementation Status section to the mutual TLS draft

kanywst:
An Implementation Status entry for `wimsey` (<https://github.com/kanywst/wimsey>), which issues and verifies Workload Identity Certificates. The document has no Implementation Status section yet, so this adds one, worded as in draft-ietf-wimse-http-signature.

The coverage line states the limits: verification is a single-issuer model against a directly provided CA, with no chain building, and server name validation is left to the caller.

`wimsey` is already listed in draft-ietf-wimse-workload-creds for the WIC itself; this entry is about the mutual TLS binding.


kanywst:
Closing this one too: the same proposal for the WPT draft (#310) drew a preference against adding an Implementation Status section, and that applies here as well. Happy to contribute an entry if a section is added later.

